PR #69, opened 2025-07-17
Zarrquet implementation: Adding Zarrs based TableProvider with bindings in Python.

alxmrs:


src/table_provider.rs
```diff
@@ -0,0 +1,1788 @@
...
+                let shape = array.shape().to_vec();
+                all_arrays.push((path_str, array, shape));
+            }
+        }
+        
+        if all_arrays.is_empty() {
+            return Err(DataFusionError::External("No arrays found in Zarr store".into()));
+        }
+        
+        // Identify data variables (highest dimensionality, >1D)
```
alxmrs, line 491: I don't think this is the definition of a data variable.

src/table_provider.rs
```diff
@@ -0,0 +1,2007 @@
...
+    type ArrowArray = Float64Array;
+    
+    fn to_arrow_array(
+        flat_data: &ndarray::ArrayBase<ndarray::CowRepr<'_, Self>, ndarray::Dim<[usize; 1]>>,
+    ) -> Vec<Self> {
+        if flat_data.is_standard_layout() {
+            flat_data.as_slice().unwrap().to_vec()
+        } else {
+            flat_data.iter().cloned().collect()
+        }
```
kylebarron, line 117: fyi I think this will never be zero-copy. Essentially any time you're calling `to_vec` on a slice it'll copy it to a new Vec

alxmrs (commented):
First part of a self-code review (for the benefit of Claude). Will resume tomorrow.

PR #70, opened 2025-07-24
Apply clippy suggestions and rustfmt to #69

kylebarron:
This adds clippy linting and rustfmt formatting to the zarr-rs branch in #69

The CI comes from https://github.com/geoarrow/geoarrow-rs/blob/main/.github/workflows/ci.yml. This version uses nightly clippy to enable import sorting. If that's not a priority for you, we can simplify the CI script to use stable clippy.

### DIFF
--- a/.github/workflows/ci-rust.yml
+++ b/.github/workflows/ci-rust.yml
@@ -1,0 +1,46 @@
+name: Rust
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


kylebarron:
You can use anything with pre-commit, though it's not very common to use pre-commit with Rust projects. E.g. in my projects I only use pre-commit for the Python parts of the Python bindings. But people have made adapters to use pre-commit with Rust https://github.com/doublify/pre-commit-rust. You might want to search for one that's maintained.

alxmrs:
Thanks for that context! In that case, this PR looks good to me.

PR #71, opened 2025-07-24
Generic conversion of array data to Arrow

kylebarron:
Some slightly simpler code for converting ndarray data to ArrayRefs by using generics.

This is on top of #70 and we should merge that first.

(I'd be in favor of merging #69 to main and then making PRs into main)


alxmrs:
Awesome, let me clean up #69 enough to land it. I'm happy to unblock you and your explorations.

PR #88, opened 2025-11-14
feat(protocol)

RohanDisa:
Set up the dataframe interchange protocol scaffolding and implemented the initial XarrayBuffer component.

Introduced a dedicated xarray_sql.dataframe_protocol subpackage so protocol-specific logic has its own organized namespace rather than crowding existing modules. Open to restructuring this if you prefer the classes to live directly under the main package.

Currently placed both the abstract Buffer interface and the concrete XarrayBuffer implementation in core.py; happy to split these into separate files if that aligns better with the project structure.

Once the layout is settled, I’ll proceed with implementing the remaining two protocol classes as discussed.

### DIFF
--- a/xarray_sql/dataframe_protocol/accessor.py
+++ b/xarray_sql/dataframe_protocol/accessor.py
@@ -1,0 +1,32 @@
+"""Dataset accessor for the DataFrame interchange protocol."""
+
+from __future__ import annotations
+
+import typing as t
+
+import xarray as xr
+
+from . import core
+
+
+AccessorReturn = t.TypeVar("AccessorReturn")
+
+
+@xr.register_dataset_accessor("df_protocol")
+class DataFrameProtocolAccessor:
+  """Access DataFrame interchange utilities from an xarray Dataset."""
+
+  def __init__(self, dataset: xr.Dataset) -> None:
+    self._dataset = dataset
+
+  def __dataframe__(self, allow_copy: bool = False):
+    """Expose the dataset via the DataFrame interchange protocol."""
+    return core.dataset_to_protocol(self._dataset, allow_copy=allow_copy)
+
+
+def register_dataset_dataframe_accessor() -> None:
+  """Ensure the dataset accessor is registered (import side-effect)."""
+  # Accessing the class ensures module import registers the accessor.
+  _ = DataFrameProtocolAccessor
+
+
--- a/xarray_sql/dataframe_protocol/accessor.py
+++ b/xarray_sql/dataframe_protocol/accessor.py
@@ -22,6 +22,3 @@ class DataFrameProtocolAccessor:
   def __dataframe__(self, allow_copy: bool = False):
     """Expose the dataset via the DataFrame interchange protocol."""
     return core.dataset_to_protocol(self._dataset, allow_copy=allow_copy)
-
-
-
--- a/xarray_sql/dataframe_protocol/accessor.py
+++ b/xarray_sql/dataframe_protocol/accessor.py
@@ -12,7 +12,7 @@ from . import core
 AccessorReturn = t.TypeVar("AccessorReturn")
 
 
-@xr.register_dataset_accessor("df_protocol")
+@xr.register_dataset_accessor("xql")
 class DataFrameProtocolAccessor:
   """Access DataFrame interchange utilities from an xarray Dataset."""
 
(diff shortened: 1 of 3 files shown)


xarray_sql/dataframe_protocol/accessor.py
```diff
@@ -0,0 +1,32 @@
...
+
+  def __init__(self, dataset: xr.Dataset) -> None:
+    self._dataset = dataset
+
+  def __dataframe__(self, allow_copy: bool = False):
+    """Expose the dataset via the DataFrame interchange protocol."""
+    return core.dataset_to_protocol(self._dataset, allow_copy=allow_copy)
+
+
+def register_dataset_dataframe_accessor() -> None:
```
alxmrs: 🐑 (the sheep emoji means a sheepish, ie. optional, suggestion). Any ideas for a more succinct name for this method? Further, how will this be used?

Maybe `register_dataframe_interchange()`? Users will have some initial context given the full import path to this method (to imply this applies to xarray datasets).

xarray_sql/dataframe_protocol/core.py
```diff
@@ -0,0 +1,205 @@
...
+  def __dlpack_device__(self) -> t.Tuple[DlpackDeviceType, t.Optional[int]]:
+    """
+    Device type and device ID for where the data in the buffer resides.
+    Uses device type codes matching DLPack.
+    Note: must be implemented even if ``__dlpack__`` is not.
+    """
+    ...
+
+
+class XarrayBuffer(Buffer):
```
alxmrs: I like this idea to subclass the buffer.

xarray_sql/dataframe_protocol/core.py
```diff
@@ -0,0 +1,205 @@
...
+  """Buffer implementation wrapping a NumPy ndarray without copying."""
+
+  def __init__(self, array: np.ndarray) -> None:
+    if not array.flags["C_CONTIGUOUS"]:
```
alxmrs: Could Zero copy transfer occur with Fortran contiguous ordering? Or is this a hard restriction?

On the Xarray ecosystem side, it's common to include these byte ordering details in an `encoding` attr. For Zarr specifically, users can configure the C or Fortran ordering, as well as the endien-ness of bytes. See the numcodecs package.

alxmrs (commented):
A closer look, this is quite promising! Thanks for your quick, quality work here.

xarray_sql/dataframe_protocol/core.py
RohanDisa: Yes, so zero copy is possible with Fortran contiguous ordering. I have made the changes.

xarray_sql/dataframe_protocol/accessor.py
alxmrs: That sounds good to me!

xarray_sql/dataframe_protocol/core.py
```diff
@@ -0,0 +1,530 @@
...
+class XarrayColumn(Column):
+  """
+  Minimal concrete Column implementation backed by an xarray.DataArray.
+
+  Assumes a 1-D DataArray representing a single logical column. If the
+  underlying array is chunked (e.g., a dask-backed DataArray), this class
+  can expose protocol chunks via num_chunks()/get_chunks().
+
+  Notes:
+    - For lazy (dask) arrays, materializing buffers requires compute. This can
```
alxmrs: I like this design choice.

xarray_sql/dataframe_protocol/core.py
```diff
@@ -0,0 +1,530 @@
...
+  """
+
+  def __init__(
+      self,
+      dataarray: xr.DataArray,
+      *,
+      allow_compute: bool = True,
+      base_offset: int = 0,
+  ) -> None:
+    if dataarray.ndim != 1:
```
alxmrs: Forgive me if I misunderstand; I don't think we want this assert. The trick we exploit in this library is that it is cheap to flatten ndarrays into vectors (a matter of changing metadata/strides). Doing this while handling chunks will take some care, but I believe it is possible.

alxmrs: It may be of help to see the "core.py" at the package root to see what I've tried in the past to achieve this effect.

alxmrs (commented):
I have concerns for the current data model of XarrayColumn, but the interfaces are looking sharp.

alxmrs:
Closing, we found another, better approach.

Issue #93, opened 2025-12-03
Does xarray-sql load datasets lazily


ahuang11:
Based on:
`ctx.from_dataset('air', ds, chunks=dict(time=24))  # the dataset needs to be chunked!`

It seems to be lazily loaded, but I wanted to confirm.

alxmrs:
I just read the sources from the pyarrow side of the dataframe interchange. Alas, I am skeptical if after this implementation, that loading xarray-sql will be lazy: https://arrow.apache.org/docs/_modules/pyarrow/interchange/from_dataframe.html#from_dataframe.

PR #95, opened 2025-12-20
Experiment: return PyArrow Dataset from read_xarray

RohanDisa:
This PR experiments with returning a PyArrow Dataset from read_xarray instead of a RecordBatchReader.

The goal is to explore whether registering chunked InMemoryDatasets improves or enables more lazy behavior when integrating with Arrow/DataFusion, in light of the ongoing discussion around dataset registration and laziness.
Issue: #93


alxmrs:
Closing this has gotten stale.

PR #100, opened 2026-01-15
Lazy datafusion registration via custom table provider

alxmrs:
Fixes #17, #93.

PR #101, opened 2026-01-16
New approach: python only stream API.

alxmrs:
TODO(alxmrs): merge into #100.


alxmrs:
Integrating into #100.

PR #113, opened 2026-02-09
Xarray-specific filter predicate pushdown.

alxmrs:
Partition pruning: In the Xarray data model, we can always get statistics for the dataset dims, which function like the "primary keys" of the materialized table. These statistics lend themselves to omit processing data that we know we don't need to. For example, 

```sql
SELECT * FROM air WHERE time > '2020-02-01'
```

Will skip loading partitions whose time ranges are entirely before 2020-02-01.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,7 +44,7 @@
 //! Supported operators: `=`, `<`, `>`, `<=`, `>=`, `BETWEEN`, `IN`, `AND`, `OR`.
 
 use std::any::Any;
-use std::collections::HashMap;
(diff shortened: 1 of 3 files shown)


alxmrs:
I'm happy to get your review when you can provide it, @maximedion2. I probably will go ahead and merge now. 

> what is a partition in this context?

A partition is one chunk of xarray data across multiple variables. Since the ideal chunks are the underlying physical chunks from the Zarr store, I've called these "blocks", to borrow from the Zarr library parlance. 

Happy to talk more about it in the future. Thanks for taking a look.

PR #141, opened 2026-03-05
Implemented Zarr / Xarray Catalog Provider for Multiple Tables

### LINKED ISSUE #85: Implementation idea: make a custom catalog provider
https://github.com/apache/datafusion-python/blob/main/python/tests/test_catalog.py

https://datafusion.apache.org/python/user-guide/data-sources.html#user-defined-catalog-and-schema

yc-606:
fixes: #85


alxmrs (changes requested):
A high level design note: I am reading more about how catalog providers work, and I am starting to think that we may not need to subclass anything from the catalog providers at all. Check out [this page of documentation](https://datafusion.apache.org/python/autoapi/datafusion/catalog/index.html). It says that constructors are not typically called directly. Furthermore, the [more I read about what CatalogProviders are for](https://datafusion.apache.org/library-user-guide/catalogs.html), the more I am beginning to think that they may not be the appropriate choice for what we really aim to do, which is register multiple tables for a single Xarray Dataset when needed. 

This issue is mainly on me, because I did not fully understand what Catalogs were for when I originally wrote up the issue. For this, I apologize. For now, what I ideally would like to see in the next revision is a simplification of the theory of this feature. For now, what if we simply altered the behavior of the `XarrayContext.from_dataset()` method so that it registered multiple tables directly? In this approach, we could reuse the core logic that you two have implemented here (grouping by dimensions, naming tables, registering tables, etc.) for making multiple tables from a single Xarray Dataset. In the new version of the method, we'd have to think of names for each of the new tables that we'd register (e.g. the dataset could be called `era5` and the subtable underneath it could be `lat_lon_time` for a full path of `era5.lat_lon_time`. Maybe we could give users the ability to override this to call these dimensions `era5.surface` instead?). 

Thanks for your hard work here, Evan and Yagna.

alxmrs:
This landed later on in #141. Closing.

PR #149, opened 2026-03-19
Add cftime support

ahuang11:
Closes https://github.com/alxmrs/xarray-sql/issues/144

After attempting this, I realized it did not feel like a good first issue~ Note I used Claude Opus to plan and write this.

Two tiers based on calendar type:
1. Gregorian-like (standard, noleap, all_leap) → pa.timestamp('us'). String SQL filters work naturally (WHERE time > '1980-01-01'). Microsecond resolution covers ±292k years, avoiding the nanosecond 1678–2262 overflow.
2. Non-Gregorian (360_day, julian) → pa.int64() with xarray:units/xarray:calendar metadata on the Arrow field. Lossless — a 360-day "Feb 30th" isn't forced into Gregorian. Auto-registered cftime() UDF provides ergonomic filtering: WHERE time > cftime('2000-07-01').

Now this runs:
```
import xarray as xr
from metpy.cbook import get_test_data

ds = xr.tutorial.open_dataset("rasm")
context = XarrayContext().from_dataset("rasm", ds, chunks={"time": 2})
context.sql("SELECT * FROM rasm LIMIT 1000")
```

image

However as you can see, there are a lot of NaNs; to be fixed in a separate PR in https://github.com/alxmrs/xarray-sql/pull/151

### DIFF
--- a/xarray_sql/df.py
+++ b/xarray_sql/df.py
@@ -384,14 +384,15 @@ def _block_metadata(coord_arrays: dict, block: Block) -> PartitionBounds:
     # (Int64/Float64/TimestampNanos) and numpy min/max ufuncs do not
     # support them.  Skip so pruning treats the dimension conservatively.
     if coord_values.dtype.kind in ("U", "S", "O"):
-      continue  # Use actual min/max rather than first/last so that non-monotonic
-      # coordinate axes (e.g. descending latitude 90→-90) are handled
-      # correctly.  np.min/max work for both numeric and datetime64 arrays.
+      continue
 
     if cft.is_cftime(coord_values):
       ranges[str(dim)] = cft.partition_bounds(coord_values)
       continue
 
+    # Use actual min/max rather than first/last so that non-monotonic
+    # coordinate axes (e.g. descending latitude 90→-90) are handled
+    # correctly.  np.min/max work for both numeric and datetime64 arrays.
     min_val = coord_values.min()
     max_val = coord_values.max()
 


xarray_sql/__init__.py
```diff
@@ -1,8 +1,10 @@
+from . import cft
 from .reader import read_xarray, read_xarray_table
 from .sql import XarrayContext
 from .df import from_map
 
 __all__ = [
+    "cft",
```
alxmrs: should we abbreviate this, or should it just also be called `cftime`?

xarray_sql/cft.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+def is_cftime_index(ds: xr.Dataset, coord_name: str) -> bool:
+  """Check if a coordinate uses a ``CFTimeIndex`` without materializing data."""
+  try:
+    idx = ds.indexes.get(coord_name)
+    if idx is not None:
+      from xarray import CFTimeIndex
+      return isinstance(idx, CFTimeIndex)
+  except Exception:
```
alxmrs: What type of exception do we expect here? I do recommend not catching a blanket exception if possible; this could swallow a useful error.

xarray_sql/cft.py
```diff
@@ -0,0 +1,194 @@
...
+  try:
+    idx = ds.indexes.get(coord_name)
+    if idx is not None:
+      from xarray import CFTimeIndex
+      if isinstance(idx, CFTimeIndex):
+        return idx.calendar  # type: ignore[attr-defined]
+  except Exception:
+    pass
+  try:
+    values = ds.coords[coord_name].values
```
alxmrs: Optional, but in other parts of this library, we "hoist" coordiante access up to a top level function. I wonder if we could simply point to that materialized data which we need to compute anyway to parse this value.

PR #151, opened 2026-03-19
Fix null conversion

ahuang11:
I noticed that the values were all NaN. This fixes it. Used Claude Opus during planning and execution.

### DIFF
--- a/tests/test_sql.py
+++ b/tests/test_sql.py
@@ -1,6 +1,7 @@
 """SQL functionality tests for xarray-sql using pytest."""
 
 import numpy as np
+import pandas as pd
 import pytest
 import xarray as xr
 
@@ -153,9 +154,6 @@ class TestNanAsNull:
 
   @pytest.fixture
   def nan_ds(self):
-    import numpy as np
-    import pandas as pd
-
     data = np.array([[[1.0, 2.0], [np.nan, 4.0]], [[5.0, np.nan], [7.0, 8.0]]])
     return xr.Dataset(
         {"temp": (["time", "x", "y"], data)},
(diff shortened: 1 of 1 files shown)


xarray_sql/df.py
```diff
@@ -204,7 +204,11 @@ def dataset_to_record_batch(
       arrays.append(pa.array(arr, type=field.type))
     else:
       # Data variable: ravel to 1-D (zero-copy for C-contiguous arrays).
-      arrays.append(pa.array(ds[name].values.ravel(), type=field.type))
+      # from_pandas=True maps NaN → Arrow null inside the C++ copy kernel,
+      # so SQL aggregates (MAX, MIN, AVG) skip missing values correctly.
+      arrays.append(
+          pa.array(ds[name].values.ravel(), type=field.type, from_pandas=True)
+      )
```
alxmrs: 🔥

xarray_sql/sql_test.py
```diff
@@ -192,3 +192,65 @@ def test_cross_join(air_and_stations):
...
+
+  def test_is_null_matches_nan(self, nan_ds):
+    ctx = XarrayContext()
+    ctx.from_dataset("data", nan_ds)
+    result = ctx.sql(
+        "SELECT COUNT(*) AS cnt FROM data WHERE temp IS NULL"
+    ).to_pandas()
+    assert result["cnt"].iloc[0] == 2
+
+  def test_from_dataset_returns_self(self):
```
alxmrs: I don't think we need this test, better to omit it for now.

Also, I think this tests behavior that is unrelated to this fix? I'm not sure I agree with this interface, as it breaks the conventions in datafusion IUUC.

alxmrs (changes requested):
Tests need some improvements. In general, thank you for making this fix!

PR #165, opened 2026-04-14
Add cloud coverage example with description and plot

eparkle:
- Added maritime sulfur restriction cloud coverage impact example to examples.md
- Added plot output png to assets (used in examples.md with cloud coverage example)


docs/examples.md
```diff
@@ -21,3 +21,77 @@ result = ctx.sql('''
...
+# ERA5 longitudes are 0–360 by default; remap to -180–180 for intuitive
+# slicing and align with standard geographic conventions
+ds = ds.assign_coords(
+    longitude = (((ds.longitude + 180) % 360) - 180)
+).sortby('longitude')
+
+# Slice North America + North Atlantic for both years
+# 2019 = pre-policy baseline; 2025 = post-IMO 2020 sulfur restriction
+ds_2019 = ds.sel(latitude=slice(75, 10), longitude=slice(-170, -50), valid_time=slice('2019-01-01', '2019-12-31'))
+ds_2025 = ds.sel(latitude=slice(75, 10), longitude=slice(-170, -50), valid_time=slice('2025-01-01', '2025-12-31'))
```
alxmrs: This example is excellent! One easy means of improvement is that we could probably move this logic into the SQL statement as a WHERE clause.

docs/examples.md
```diff
@@ -21,3 +21,77 @@ result = ctx.sql('''
...
+between 2019 (pre-policy baseline) and 2025 (post-IMO 2020 low-sulfur fuel mandate),
+using ERA5 reanalysis data queried with xarray-sql.
+
+![Monthly cloud coverage comparison 2019 vs 2025](assets/cloudcoveragegraph.png)
+```python
+import xarray as xr
+import xarray_sql as xql
+import matplotlib.pyplot as plt
+
+ds = xr.open_dataset("era5_dataset.nc", chunks='auto')
```
alxmrs: Where does this netcdf file come from? I bet we could replace this local file with ARCO-ERA5 (a Zarr dataset) and it would still work. 

https://github.com/google-research/arco-era5

Cargo.lock
```diff
@@ -3375,7 +3375,7 @@ checksum = "ea2f10b9bb0928dfb1b42b65e1f9e36f7f54dbdf08457afefb38afcdec4fa2bb"
 
 [[package]]
 name = "xarray_sql"
-version = "0.2.1"
+version = "0.2.2"
```
alxmrs, line 3378: Let's not change this version until we do a release -- unless, is this the same version as main?

alxmrs (commented):
Thank you so much for submitting your example to the project!

alxmrs:
Hey @eparkle, would you like to update these examples based on my feedback? I do really like this idea and it would be welcome in the project.

PR #166, opened 2026-04-29
Made new xarray context behavior to register multiple tables from a dataset

eparkle:
This should account for the differing dimensions between variables

### DIFF
--- a/xarray_sql/sql.py
+++ b/xarray_sql/sql.py
@@ -13,32 +13,32 @@ class XarrayContext(SessionContext):
 
     def from_dataset(
         self,
-        table_name: str,
+        name: str,
         input_table: xr.Dataset,
         *,
-        dim_group_aliases: dict[tuple[str, ...], str] | None = None,
+        table_names: dict[tuple[str, ...], str] | None = None,
         chunks: Chunks = None,
     ):
         """Register an xarray Dataset as one or more queryable SQL tables.
 
         When all data variables share the same dimensions, the dataset is
-        registered as a single table named ``table_name``. When variables
-        have differing dimensions (e.g. some on a 3D grid and others on a
-        4D grid), the dataset is split into one table per dimension group.
+        registered as a single table named ``name``. When variables have
+        differing dimensions (e.g. some on a 3D grid and others on a 4D
+        grid), the dataset is split into one table per dimension group.
         The tables are registered under a SQL schema (namespace) named
-        ``table_name`` and named ``<dim1>_<dim2>_...`` by default::
+        ``name`` and named ``<dim1>_<dim2>_...`` by default::
 
             ctx.from_dataset('era5', ds, chunks={'time': 24})
             # registers tables: 'era5.time_lat_lon' and
             #                   'era5.time_lat_lon_level'
             ctx.sql('SELECT AVG(temperature_2m) FROM era5.time_lat_lon')
 
-        Use ``dim_group_aliases`` to override the suffix for specific
-        dimension tuples::
+        Use ``table_names`` to override the name for specific dimension
+        tuples::
 
             ctx.from_dataset(
                 'era5', ds,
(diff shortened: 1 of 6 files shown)


xarray_sql/sql.py
```diff
@@ -61,3 +63,61 @@ def from_dataset(
...
+        for dims, var_names in groups.items():
+            suffix = dim_aliases.get(dims, "_".join(dims))
+            table_name = f"{dataset_name}_{suffix}"
+            sub_ds = input_table[var_names]
+            self.from_dataset(table_name, sub_ds, chunks)
+
+        return self
+
+
+def _group_vars_by_dims(ds: xr.Dataset) -> dict[tuple[str, ...], list[str]]:
```
alxmrs, line 119: I'd like to see a few unit tests for this method, even though it is private.

xarray_sql/sql.py
```diff
@@ -61,3 +63,61 @@ def from_dataset(
                     break  # One UDF per context is enough.
 
         return self
+
+    def from_dataset_multi(
```
alxmrs: High level feedback: I think we should roll this into `from_dataset` and it should be the "default" behavior. Since we are an experimental project, I'm OK with changing the public API. 

I think @ghostiee-11 or @ahuang11  might have a few ideas about what good names would be for these kwargs, i.e. what is most conventional for Xarray. So far, these LGTM.

xarray_sql/sql.py
```diff
@@ -61,3 +63,61 @@ def from_dataset(
...
+        self,
+        dataset_name: str,
```
alxmrs: This should be `table_name` IMO

xarray_sql/sql.py
```diff
@@ -61,3 +63,61 @@ def from_dataset(
...
+        input_table: xr.Dataset,
+        dim_aliases: dict[tuple[str, ...], str] | None = None,
```
alxmrs: I think should be a keyword only argument, see: https://python-3-for-scientists.readthedocs.io/en/latest/python3_advanced.html

alxmrs: I like "aliases", but "dim" is a bit unclear; I'm not sure exactly what these should be called. Maybe "dim_group_aliases"? I'm curious what @ahuang11 and @ghostiee-11  think, too.

xarray_sql/sql.py
```diff
@@ -61,3 +63,61 @@ def from_dataset(
...
+
+            ctx.from_dataset_multi('era5', ds, chunks={'time': 24})
+            # registers: 'era5.time_lat_lon' and 'era5.time_lat_lon_level'
+            ctx.sql('SELECT AVG(temperature_2m) FROM "era5.time_lat_lon"')
+
+        Args:
+        dataset_name: A name for the dataset, used as a prefix for all
+            registered table names.
+        input_table: An xarray Dataset. Variables may have differing dimensions.
+        dim_aliases: Optional mapping from dimension tuples to custom table
```
alxmrs: I like that we have an aliases override.

xarray_sql/sql.py
```diff
@@ -61,3 +63,61 @@ def from_dataset(
...
+        """
+
+        dim_aliases = dim_aliases or {}
+        groups = _group_vars_by_dims(input_table)
+
+        for dims, var_names in groups.items():
+            suffix = dim_aliases.get(dims, "_".join(dims))
+            table_name = f"{dataset_name}_{suffix}"
+            sub_ds = input_table[var_names]
+            self.from_dataset(table_name, sub_ds, chunks)
```
alxmrs: I like this pattern, so much so that I think in the new version that we could turn the existing `from_dataset` into a private method (`_from_dataset`). Then, in a method like this, we could have a big `if` or return early logic to tell if there is more than one group of dims.

alxmrs (commented):
Looking really good! Here are my thoughts on design.

PR #167, opened 2026-05-13
Add lazy SQL → xarray round-trip via XarrayDataFrame.to_dataset (closes #58)

### LINKED ISSUE #58: Inverse problem: Read a table or SQL query into an Xarray Dataset
I had a shower thought today. If Zarr is an OLAP DB, then could we treat tables as rasters?

https://github.com/zarr-developers/zarr-specs/issues/290

One perspective is that this is the inverse of what this project is trying to do so far, which is to treat rasters as tables to run SQL queries against them. 

One fuzzy idea for a use case like this would be to read a table of data associated with a latlng and translate it to a (sparse) raster.

Similar to #4, I wonder if column-oriented data like parquet, which most headless tabular data stores are represented today, would naturally fit well with Xarray-style chunks.

ghostiee-11:
## Summary

Closes #58. `ctx.sql(...)` now returns an `XarrayDataFrame` that adds `.to_dataset()` on top of the existing `.to_pandas()`. The Dataset is lazy by default slicing pushes down into SQL `WHERE` clauses so only the requested slab gets materialized.

## Why

Going `Dataset → SQL` worked great. Going the other way meant hand-rolling `set_index().to_xarray()` and waving goodbye to attrs, encoding, and non-dim coords. This closes the loop. API shape follows the wrapper proposal in the issue thread; lazy-default via `BackendArray` + `LazilyIndexedArray` is the direction that came out of the design discussion.

## Before / After

**Before**
```python
df = ctx.sql('SELECT * FROM "air"').to_pandas()        # +194 MB RSS
ds = df.set_index(['time', 'lat', 'lon']).to_xarray()  # attrs / encoding / non-dim coords gone
```

**After**
```python
out  = ctx.sql('SELECT * FROM "air"').to_dataset()     # +2 MB RSS, lazy
slab = out['air'].isel(time=0).values                  # +15 MB, single wrapped SQL with WHERE time=...
# attrs, encoding (dtype-bound keys stripped), non-dim coords, dim dtype all back
```

RSS on `xr.tutorial.open_dataset('air_temperature')` (~31 MB):

| Path | RSS delta |
|---|---|
| `ctx.sql(...).to_pandas()` | **+194 MB** |
| Hand-written `WHERE` query | +18 MB |
| **Lazy `to_dataset()` + `isel(time=0).values`** | **+17 MB** |

About 11× lighter than the eager path, basically tied with hand-written filtered SQL.

## Scope

- `xarray_sql/ds.py`: `XarrayDataFrame` + `SQLBackendArray` (pushes indexers down to SQL `WHERE`) + template-based metadata recovery
- `XarrayContext.sql()` returns the wrapper; `.to_pandas()` stays bit-for-bit identical (there's a regression test for this)
- Lazy via `BackendArray` + `LazilyIndexedArray` with `IndexingSupport.OUTER`; dask stays orthogonal
- `sparse_extent="result"|"template"` for filtered queries
- 48 new tests, README example

Follow-ups per the issue thread: chunking / rechunker primitive, hypothesis property tests, simulation testing, promoting the standalone helper to public.

## Test plan

- [x] `uv run pytest tests/` - 158/158
- [x] `uv run ruff check xarray_sql/ tests/` clean
- [x] `uv run ruff format -check xarray_sql/ tests/` clean
- [x] `uv run python -m mypy xarray_sql/` - 4 errors, all pre-existing (net −1 vs main)
- [x] README example runs end-to-end


tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
+"""Tests for the SQL -> xarray reverse path (xarray-sql#58).
```
alxmrs: Let's remove the issue number from the docstring. Claude has a habit of including text like this, but it doesn't make sense outside of the context of the conversation.

tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
...
+
+Phase 1 covers the eager round-trip:
+
+* :class:`xarray_sql.XarrayDataFrame` returned by ``ctx.sql``
+* ``.to_dataset()`` with explicit and auto-inferred ``dim_cols``
+* Template-based metadata recovery (var attrs/encoding, dataset attrs,
+  non-dim coords, dim-coord dtype)
+* FROM-clause regex unit tests
+
+Lazy semantics land in Phase 2; sparse extent and edge cases in Phase 3.
```
alxmrs: Similarly, I don't know what the phases are; to my knowledge they aren't documented anywhere (nor should they be IMO).

tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
...
+    result = ctx.sql("SELECT * FROM air LIMIT 5")
+    assert isinstance(result, XarrayDataFrame)
+
+
+def test_to_pandas_unchanged_behavior(air_dataset_small):
+    """Wrapped ``.to_pandas()`` is bit-for-bit equal to the un-wrapped path."""
+    ctx = XarrayContext()
+    ctx.from_dataset("air", air_dataset_small)
+    wrapped = ctx.sql("SELECT * FROM air LIMIT 7").to_pandas()
+    raw = super(type(ctx), ctx).sql("SELECT * FROM air LIMIT 7").to_pandas()
```
alxmrs: Why do we do `super(type(ctx), ctx)`?

tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
...
+    pd.testing.assert_frame_equal(wrapped, raw)
+
+
+def test_passthrough_methods(air_dataset_small):
```
alxmrs, line 51: I'm not sure the purpose of this test. Does this test an internal/private impl detail, or is this public facing?

tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
...
+# Round-trip via to_dataset (explicit dim_cols)
+# ---------------------------------------------------------------------------
+
+
+def test_to_dataset_explicit_dims_select_star(air_dataset_small):
+    ctx = XarrayContext()
+    ctx.from_dataset("air", air_dataset_small)
+    out = ctx.sql("SELECT * FROM air").to_dataset(
+        dim_cols=["time", "lat", "lon"]
+    )
```
alxmrs: Why does the user have to set `dim_cols` at all? Isn't that parsable from the Xarray dataset?

tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
...
+def test_round_trip_select_star_values_match(air_dataset_small):
+    """Values survive the round-trip (modulo ascending coord ordering)."""
+    ctx = XarrayContext()
+    ctx.from_dataset("air", air_dataset_small)
+    out = ctx.sql("SELECT * FROM air").to_dataset(
+        dim_cols=["time", "lat", "lon"]
+    )
+    expected = air_dataset_small.compute().sortby(["time", "lat", "lon"])
+    actual = out.sortby(["time", "lat", "lon"])
+    np.testing.assert_array_equal(actual["air"].values, expected["air"].values)
```
alxmrs: Can we also add an xarray level equality to test that the dataframes (i.e., which also have metadata) are equal?

tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
...
+    )
+    assert isinstance(out, xr.Dataset)
+    assert set(out.dims) == {"time", "lat", "lon"}
+    assert "air" in out.data_vars
+    assert out.sizes["lat"] == air_dataset_small.sizes["lat"]
+    assert out.sizes["lon"] == air_dataset_small.sizes["lon"]
+    assert out.sizes["time"] == air_dataset_small.sizes["time"]
+
+
+def test_round_trip_select_star_values_match(air_dataset_small):
```
alxmrs: Let's use pytest to set up a fixture that will return all the xarray tutorial datasets -- this way, we can check a wider variance of datasets via a rountrip property. Later, hypothesis tests can come.

tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
...
+    ctx.from_dataset("air", air_dataset_small)
+    out = ctx.sql("SELECT * FROM air").to_dataset(
+        dim_cols=["time", "lat", "lon"]
+    )
+    expected = air_dataset_small.compute().sortby(["time", "lat", "lon"])
+    actual = out.sortby(["time", "lat", "lon"])
+    np.testing.assert_array_equal(actual["air"].values, expected["air"].values)
+
+
+def test_round_trip_preserves_dim_order(air_dataset_small):
```
alxmrs: Can we combine this test with the above test?

tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
...
+    ctx.from_dataset("air", air_dataset_small)
+    out = ctx.sql(
+        "SELECT lat, lon, AVG(air) AS air_avg FROM air GROUP BY lat, lon"
+    ).to_dataset(dim_cols=["lat", "lon"])
+    assert set(out.dims) == {"lat", "lon"}
+    assert "air_avg" in out.data_vars
+    assert "air" not in out.data_vars
+    expected = (
+        air_dataset_small.compute()
+        .sortby(["lat", "lon"])
```
alxmrs, line 134: Do we need to sort these? CC: @jayendra13 might be working on something similar here.

tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
...
+    assert out["air"].dims == expected_dims
+
+
+def test_aggregation_drops_dim(air_dataset_small):
+    """``GROUP BY lat, lon`` over time -> 2D Dataset with the alias."""
+    ctx = XarrayContext()
+    ctx.from_dataset("air", air_dataset_small)
+    out = ctx.sql(
+        "SELECT lat, lon, AVG(air) AS air_avg FROM air GROUP BY lat, lon"
+    ).to_dataset(dim_cols=["lat", "lon"])
```
alxmrs: I see the value of `dim_cols` here, given that we are "remapping" dimensions.

alxmrs: I don't like the term "dim_cols" -- these are two abbreviations. I'd prefer if we used some sort of term from the DataFusion ecosystem. I typically call these "primary_keys_columns" (that is what they are IMO). This may be confusing for Xarray users; I'm open to naming ideas.

tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
...
+    assert "air_avg" in out.data_vars
+    assert "air" not in out.data_vars
+    expected = (
+        air_dataset_small.compute()
+        .sortby(["lat", "lon"])
+        .mean(dim="time")["air"]
+        .values
+    )
+    actual = out.sortby(["lat", "lon"])["air_avg"].values
+    np.testing.assert_allclose(actual, expected)
```
alxmrs, line 139: let's see if we can use xarray's equality system instead of mapping to numpy -- maybe this would omit the need to sort. 

https://docs.xarray.dev/en/stable/generated/xarray.testing.assert_equal.html

alxmrs (changes requested):
First pass at a review. I have some high level feedback interface feedback implied here, and then notes on how the tests are written.

tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
...
+
+
+# ---------------------------------------------------------------------------
+# dim_cols inference
+# ---------------------------------------------------------------------------
+
+
+def test_to_dataset_infers_dim_cols_from_single_registration(
```
alxmrs: This test seems redundant. Isn't this captured elsewhere? In general, when using LLM coding assistants, it's important to keep tests high signal verification channels.

tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
...
+            "SELECT lat, lon, AVG(air) AS air_avg FROM air GROUP BY lat, lon"
+        ).to_dataset()
+
+
+# ---------------------------------------------------------------------------
+# Template-based metadata recovery
+# ---------------------------------------------------------------------------
+
+
+def test_template_recovers_dataset_attrs(air_dataset_small):
```
alxmrs: I think these tests could also be captured in an xarray testing equality check.

tests/test_ds.py
```diff
@@ -0,0 +1,691 @@
...
+    }
+    ctx = XarrayContext()
+    ctx.from_dataset("air", ds)
+    out = ctx.sql("SELECT * FROM air").to_dataset(
+        dim_cols=["time", "lat", "lon"]
+    )
+    assert out["air"].encoding.get("zlib") is True
+    assert "dtype" not in out["air"].encoding
+    assert "_FillValue" not in out["air"].encoding
+    assert "missing_value" not in out["air"].encoding
```
alxmrs, line 335: I'd like to know more about this API contract. Can you say more? Why is this the case? I'll keep reading to try to find out.

xarray_sql/ds.py
```diff
@@ -0,0 +1,762 @@
+"""Reverse path: tabular SQL results back to xarray Datasets.
```
alxmrs: Can we make this more like a grammatically correct sentence? This is the top level line that will show up in our auto-generated documentation.

xarray_sql/ds.py
```diff
@@ -0,0 +1,762 @@
...
+  ``xr.Dataset`` while keeping every other DataFusion method available.
+
+``.to_dataset()`` is lazy by default for ``SELECT *``-style queries: data
+variables are backed by :class:`SQLBackendArray` wrapped in
+``xarray.core.indexing.LazilyIndexedArray``. Slicing and ``.sel`` translate
+to SQL ``WHERE`` clauses pushed into a sub-query of the original SQL, so
+only the requested slab is materialized. Aggregation queries (where the
+result contains columns that are not in the registered template) fall
+back to an eager materialize-and-reshape path. ``.compute()`` always
+returns an in-memory Dataset.
```
alxmrs: Why do we "push down" the `.sel` statements into SQL? That's cool, but I am concerned that this will complicate the implementation. If we implement the `sel` operator to map it to the equivalent SQL operator (`WHERE`), why stop there? It suggests that we make an operator parity for the whole API surface. I don't think we need to do that. What do you think?

xarray_sql/ds.py
```diff
@@ -0,0 +1,762 @@
...
+import warnings
+from dataclasses import dataclass, field
+from typing import Any, Literal, cast
+
+import numpy as np
+import pandas as pd
+import pyarrow as pa
+import xarray as xr
+
+SparseExtent = Literal["result", "template"]
```
alxmrs: What is an "Extent"? I'm not sure what this is for upon first read. Can you add a comment or docstring? This is not a term of art that I'm familiar with.

xarray_sql/ds.py
```diff
@@ -0,0 +1,762 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Registry view (shared between XarrayContext and the wrapper)
+# ---------------------------------------------------------------------------
+
+
+@dataclass
+class _RegistryView:
+    """Snapshot of ``XarrayContext`` registrations handed to a wrapper.
```
alxmrs: I see the value of this class. What is a "registration"?

xarray_sql/ds.py
```diff
@@ -0,0 +1,762 @@
...
+# ---------------------------------------------------------------------------
+
+
+# Matches "FROM ident" / "JOIN ident" with bare or double-quoted identifiers.
+# Subqueries (``FROM (``) and CTE bodies do not match because ``(`` is not a
+# valid identifier start.
+_FROM_OR_JOIN_RE = re.compile(
+    r'\b(?:FROM|JOIN)\s+(?:"([^"]+)"|([A-Za-z_][A-Za-z0-9_]*))',
+    re.IGNORECASE,
+)
```
alxmrs: This is concerning to me. What if the parser in DataFusion changed when we upgraded? Would we have to re-invent this parser, too?

xarray_sql/ds.py
```diff
@@ -0,0 +1,762 @@
...
+
+# Matches a bare unfiltered SELECT *, e.g. ``SELECT * FROM "air"``. Used by
+# ``_lazy_to_xarray`` to skip the DISTINCT-per-dim coord scan when the
+# query is known to cover the full registered Dataset.
+_UNFILTERED_SELECT_STAR_RE = re.compile(
```
alxmrs: Same here.

xarray_sql/ds.py
```diff
@@ -0,0 +1,762 @@
...
+    r'^\s*SELECT\s+\*\s+FROM\s+"?([A-Za-z_][A-Za-z0-9_]*)"?\s*;?\s*$',
+    re.IGNORECASE | re.DOTALL,
+)
+
+
+def _extract_from_tables(query: str) -> list[str]:
```
alxmrs: I need to get to the "meat" of this file, but I continue to be concerned that this layer involves parsing.

alxmrs (commented):
Another round of review.

tests/test_ds.py
ghostiee-11: Sure!! Will remove this

tests/test_ds.py
ghostiee-11: Yess!! Will remove this also

tests/test_ds.py
ghostiee-11: It bypasses our sql() override to get the raw DataFusion result for an A/B compare. Will be replacing with datafusion.SessionContext.sql(ctx, ...).

tests/test_ds.py
ghostiee-11: It is public-facing, it verifies DataFusion methods we don't override `(.schema(), .collect(), .count())` still work through the wrapper's `__getattr__`, i.e. existing `ctx.sql(...)` callers are unaffected. The `repr(result) == repr(result._inner)` line does reach a private attr; dropping that line, keeping the public-method assertions.

tests/test_ds.py
ghostiee-11: It is parsed from the dataset for `SELECT *` `test_to_dataset_infers_dim_cols_from_single_registration `covers that. You only have to pass it when the query remaps dims (e.g. an aggregation drops one), since then nothing matches to infer from. This test explicitly sets it to isolate the reshape, but I'll rely on inference in the non-remapping tests, so the default path is exercised.

tests/test_ds.py
ghostiee-11: Okay!! moving the round-trip tests to xr.testing.assert_identical so values, dims, coords, dtype, and attrs all get checked together.

tests/test_ds.py
ghostiee-11: Sure!! I'll add this fixture over the tutorial datasets and a single round-trip property test across them. But, `xr.tutorial.open_dataset` pulls over the network via pooch, so if CI runs offline, this needs caching or a skip marker. Hypothesis on top of that is a natural follow-up.

tests/test_ds.py
ghostiee-11: Yeah!! Sure will fold it together

tests/test_ds.py
ghostiee-11: The sort is only needed because the lazy path emits coords ascending `(ORDER BY "d") `while sources like air come back lat-descending. Rather than sort in the test, I'll have the lazy path reindex onto the template's coord order so the round-trip is genuinely identical. Happy to sync with @jayendra13 if there's overlap here.

tests/test_ds.py
ghostiee-11: > I see the value of `dim_cols` here, given that we are "remapping" dimensions.

Correct - `GROUP BY lat, lon` drops time, so no registered Dataset's dims match the result and the caller must state them.

ghostiee-11: Okay, agreed!! I lean to dimension_columns with the docstring noting "the result's primary key."

tests/test_ds.py
ghostiee-11: You're right, this stops carrying its weight once the round-trip property test relies on inference. Removing it.

tests/test_ds.py
ghostiee-11: Agreed

tests/test_ds.py
ghostiee-11: SQL can change a column's `dtype `- `int16` becomes `float64` after an `AVG`, or after a filter introduces nulls. If we re-attach the source's encoding verbatim`(dtype=int16, _FillValue=-999)`, a later `ds.to_netcdf()` will try to pack float data back into int16 and write garbage. So we keep the safe parts of encoding (compression, chunking) and drop the dtype-bound keys. I'll put this reasoning in the docstring.

xarray_sql/ds.py
ghostiee-11: Yeah correct!! Can make it "_Reconstruct xarray Datasets from SQL query results - the inverse of the Dataset-to-table pivot."_

tests/test_ds.py
ghostiee-11: Checked the doc you linked. `assert_equal `gets us most of the way: with `check_dim_order=False` it stops caring about dim order, which kills one source of the `sortby`. It still won't reorder coordinate values within a dim though, and our lazy path emits lat ascending while air is lat-descending - so element-wise it'd still fail without a sort. The real fix is to have the lazy path reindex onto the template's coord order; then it's a clean `assert_equal` with `check_dim_order=False `and no sorting. 
One thing to note: `assert_equal ` ignores attrs, so the metadata-recovery tests will use assert_identical instead (it checks attrs and names), while the pure value/coord round-trips use `assert_equal`.

xarray_sql/ds.py
ghostiee-11: The reason it pushes `.sel ` down is that pushdown is what makes the result actually lazy, that was the core ask on the issue, and without it every `.values` would pull the whole result into memory. But I don't think it slides toward whole-API parity: the BackendArray contract is a single method,` __getitem__` with positional keys. xarray is the one that lowers `.sel/.isel/slicing` into those positional calls, we just translate them to a `WHERE` on the coord values. There's no path to `.mean/.groupby/joins` because there's only the one entry point. If you'd still rather keep it simple for v1, the alternative is materialize-on-first-access, no WHERE synthesis, but it gives up the ~11× memory win and the laziness. I'm fine going either way; your call on the tradeoff.

xarray_sql/ds.py
ghostiee-11: Yeah, "Extent" isn't a real term, that's by me. It selects a filtered query's output grid: "result" keeps only the rows the query produced; "template" reindexes back onto the source's full coords and fills the gaps. Renaming it to something else and documenting it

xarray_sql/ds.py
ghostiee-11: I just mean the `table_name` → `xr.Dataset` pairs you create with `ctx.from_dataset(...)`. I'll rename `_RegistryView` and rewrite the docstring to say that plainly instead of this.

xarray_sql/ds.py
ghostiee-11: It's not parsing SQL, and nothing about correctness depends on it. `_extract_from_tables` only picks which registered dataset to use as the metadata template when you've registered several. If it matches nothing, you pass `template_table= explicitly ', clear error, never a wrong result. So a DataFusion dialect change just pushes you to the explicit path; it never makes us own a parser. I'd rather strip this.

xarray_sql/ds.py
ghostiee-11: Same shape of thing here - this one only decides whether we can skip the per-dim `SELECT DISTINCT` coord scan for a bare `SELECT *`. No match just means we fall back to the `DISTINCT` path, which is correct, only slower. I can drop it too and always take the DISTINCT path if you'd rather not have the heuristic at all.

xarray_sql/ds.py
ghostiee-11: Fair, I'd rather not make you keep raising it. To be clear about what's actually here: two small regex heuristics, neither reads the query. `_extract_from_tables `only picks which registered dataset to use as the metadata template when several are registered (no match, you pass` template_table= yourself`, clear error, not a wrong result). `_UNFILTERED_SELECT_STAR_RE` only decides whether we can skip the per-dim `SELECT DISTINCT` scan (no match, slower DISTINCT path). Nothing about correctness leans on them.

Simplest path: I pull both out for v1. Explicit `template=/template_table=`, always DISTINCT, zero string inspection in the diff. Auto-resolution could come back later from DataFusion's own plan output rather than our regex, but I haven't checked what datafusion-python exposes there, so I'd leave that as a follow-up. Sound right to you?

ghostiee-11:
Two things I want your call on before I push the review changes, so I'm not churning the diff twice.

**1. The SQL regexes.** You flagged the parsing in a few places and I think you're right to. Options as I see them:

- Strip both for v1: explicit `template=`/`template_table=`, always the DISTINCT coord path. Zero string inspection left. Cost is a bit more typing for multi-table users and a DISTINCT scan on every lazy construct instead of the `SELECT *` fast path.
- Keep them with the degradation documented: smaller diff, keeps the construction-time optimization, but the concern you raised is still sitting there.
- Middle: drop the FROM-clause template auto-resolution (the part you were most wary of) but keep the `SELECT *` fast-path regex, which is a pure perf shortcut with no correctness role.

I lean toward stripping both, since "no parsing at all" is the cleanest answer to the concern and the perf cost is recoverable later from DataFusion's own plan output rather than a regex. But it's your tradeoff to make.

**2. The `dim_cols` name.** Agree the double abbreviation is bad. `dimension_columns` is xarray-native and what xarray users will expect; `primary_key_columns` (your suggestion) is the more honest relational name but isn't xarray vocabulary. I'd go `dimension_columns` with the docstring noting it's the result's primary key, but happy to use `primary_key_columns` if you'd rather lead with the relational framing.

Whichever you pick on both, I'll batch all the review fixes into one push.

ghostiee-11:
Hey @alxmrs !! Wanted you to have a look at this. 
I'll fix it according to your preference.

alxmrs:
Thanks for the ping; I'll try to get to this tonight!

README.md
```diff
@@ -66,6 +66,54 @@ df.head()
...
+eager = out.compute()
+```
+
+`dim_cols` defaults to the dims of the registered Dataset referenced by the
+SQL `FROM` clause. Variable attrs, dataset attrs, non-dimension coords, and
+dim-coordinate dtypes are recovered from the registered Dataset
+automatically.
+
+For filtered queries that return only part of the original extent, pass
+`sparse_extent="template"` to reindex back to the full grid with NaN
```
alxmrs: These names seem unintuitive to me, but I like that we can make sparse representations of the Dataset.

xarray_sql/ds.py
```diff
@@ -0,0 +1,762 @@
...
+        return f"'{escaped}'"
+    raise TypeError(
+        f"Cannot build SQL literal for value of type {type(value).__name__}"
+    )
+
+
+class SQLBackendArray(xr.backends.BackendArray):
+    """Lazy N-D array backed by re-executing a wrapped SQL query.
+
+    Translates xarray indexers (BasicIndexer, OuterIndexer; Vectorized
```
alxmrs: I think I see the strategy now. This is very neat. Using subqueries to get ranges is smart, though I assumed there would be some interface in DataFusion internals to accomplish the same thing (e.g. chunks via recordbatch readers).

xarray_sql/ds.py
```diff
@@ -0,0 +1,762 @@
...
+            return self._fallback_materialize_and_index(key)
+
+        cols = ", ".join(f'"{c}"' for c in self._dim_cols + [self._var_name])
+        where_clause = f" WHERE {' AND '.join(conds)}" if conds else ""
+        wrapped = (
+            f"SELECT {cols} FROM ({self._base_query}) AS _xql_base"
+            f"{where_clause}"
+        )
+        try:
+            raw_df = _raw_sql(self._ctx, wrapped).to_pandas()
```
alxmrs: I think you're using SQL and subqueries to parse things we get "for free" with the data fusion dataframe interfaces. 

I believe that the non-pandas, default df that is returned from a context.query has what we need to iterate in partitions/chunks without dipping into SQL statements and subqueries. Though, I could be wrong here. 

I ask that you do some more reading of the DataFusion docs to see if there is a simpler way.

xarray_sql/ds.py
```diff
@@ -0,0 +1,762 @@
...
+        except Exception:
+            return self._fallback_materialize_and_index(key)
+
+        if len(raw_df) == 0:
+            # WHERE matched no rows: caller asked for empty slab.
+            return cast(np.ndarray, np.empty(out_shape, dtype=self.dtype))
+
+        da = raw_df.set_index(list(self._dim_cols)).to_xarray()[self._var_name]
```
alxmrs: I was hoping our implementation wouldn't go through pandas (involves a lot of copying) and directly integrate the da/ds abstractions with arrow

alxmrs (commented):
While it seems like this works and the shape of the integration is correct (using the right Xarray BackingArray interfaces), I believe there is a simpler approach that can be done more efficiently and more elegantly that directly melds Xarray and DataFusion/Arrow.

xarray_sql/sql.py
```diff
@@ -61,3 +70,29 @@ def from_dataset(
                     break  # One UDF per context is enough.
 
         return self
+
+    def sql(self, query: str, *args, **kwargs) -> XarrayDataFrame:
```
alxmrs, line 140: This impl is solid

xarray_sql/ds.py
alxmrs: Can we make use of execute stream (or the partitioned version), for example?

https://datafusion.apache.org/python/autoapi/datafusion/dataframe/index.html#datafusion.dataframe.DataFrame.execute_stream

ghostiee-11:
Hey @alxmrs, pushed the rearchitecture in `ade544a` on top of the existing PR commits. One commit covering everything you flagged around, plus the round-1 cleanup.

**The big shifts**

- The slab access path is now `df.filter(expr).select(*cols).execute_stream()` against the wrapper's live DataFusion DataFrame, scattering Arrow RecordBatches into a preallocated numpy buffer. No SQL string subqueries, no pandas hop. Filter predicates are typed `Expr` objects (`col(d) == literal(v)`, Or-chained for IN-equivalent).
- Per-dim coord discovery in the lazy constructor is also `inner_df.select(col(d)).distinct().sort(...).execute_stream()`. Arrow-native, no DISTINCT SQL strings.
- The eager / aggregation path got the same treatment: `_to_arrow_table` consumes the wrapper via `execute_stream` (no pandas pass-through), then the same Arrow scatter helper.
- All the parsing is gone: both regexes, `_extract_from_tables`, `_sql_literal`, `_build_cond`, `_raw_sql`, and `import re` are deleted from `ds.py`.

**Round-1 cleanup folded in**

- `dim_cols` parameter renamed to `dimension_columns`.
- 8 narrow round-trip and template-recovery tests collapsed into one parametrized property test using `xr.testing.assert_identical` over `air_dataset_small`, `weather_dataset`, and `synthetic_dataset`. That one test now covers values, dims, coord values, dim-coord dtype, dim-coord attrs, non-dim coords, and dataset attrs in a single check per fixture.
- `_apply_template` now also copies dim-coord attrs (`standard_name`, `long_name`, `units`) so the round-trip is identical down to coord metadata. Found this gap when wiring up `assert_identical`.
- Module docstring on `ds.py` is a sentence; "Phase 1/2/3" and the issue ref are gone from the test docstring; `super(type(ctx), ctx)` and the `repr == _inner` private-attr line are gone.
- `SparseExtent` now has an inline comment explaining what "result" vs "template" mean. Kept the parameter name since you flagged the naming but didn't propose a replacement; happy to rename if you have a preference.

**Memory**

On `air_temperature` (~31 MB):

| Path | Pre-rewrite | Post-rewrite |
|---|---|---|
| Lazy construction | +2 MB | +2 MB |
| Lazy slab (`isel(time=0).values`) | +17 MB | **+2.3 MB** |

About 7x cheaper on the slab path. The pandas detour was most of the cost.

**Net diff**

Despite adding the Arrow scatter helper, the Arrow-native eager path, and the new parametrized property test, the diff is +492 / -543 = -51 lines.

ghostiee-11:
Hey @alxmrs, want your another round of review to the changes!!
Thankyou

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+SparseExtent = Literal["result", "template"]
+
+
+# ---------------------------------------------------------------------------
+# Registry view (shared between XarrayContext and the wrapper)
+# ---------------------------------------------------------------------------
+
+
+@dataclass
+class _RegistryView:
```
alxmrs: Now that this only holds a dictionary, could we replace this with just the dictionary? Do we have evidence that we'll need this interface later on? Let's not make the abstraction until we know; YAGNI.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+    ``ds.dims`` keys for empty Datasets. Always use this rather than
+    ``list(ds.dims)`` when round-tripping, since the latter is in
+    canonical name order and may not match the variable's axis order.
+    """
+    if ds.data_vars:
+        return list(next(iter(ds.data_vars.values())).dims)
+    return list(ds.dims)
+
+
+def _to_arrow_table(result: Any) -> pa.Table:
```
alxmrs: I recommend adding a case to cover the data interchange protocol. 

https://arrow.apache.org/docs/python/interchange_protocol.html#from-other-libraries-to-pyarrow-from-dataframe

alxmrs: In fact, it might be worthwhile to make `XarrayDataFrame` adhere to this protocol; then this function could probably be replace with `from_dataframe()`.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+        f"Unsupported result type {type(result).__name__!r}; expected "
+        "pa.Table, pd.DataFrame, datafusion.DataFrame, or XarrayDataFrame"
+    )
+
+
+def _drop_null_dim_rows_arrow(
+    table: pa.Table, dimension_columns: list[str]
+) -> pa.Table:
+    """Drop rows with null dim coords from a ``pa.Table``. Warns once."""
+    import pyarrow.compute as pc
```
alxmrs: Ideally, this would be a import at the top of the file. Would be good to ask your agent the tradeoffs for this.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+    raise TypeError(
+        f"Unsupported result type {type(result).__name__!r}; expected "
+        "pa.Table, pd.DataFrame, datafusion.DataFrame, or XarrayDataFrame"
+    )
+
+
+def _drop_null_dim_rows_arrow(
+    table: pa.Table, dimension_columns: list[str]
+) -> pa.Table:
+    """Drop rows with null dim coords from a ``pa.Table``. Warns once."""
```
alxmrs: This seems useful. Can you explain how this came up / why it is needed (for my own understanding)?

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+    ]
+    warnings.warn(
+        f"Dropping {n_dropped} row(s) with null dim values in "
+        f"columns {null_cols} before reshape",
+        stacklevel=3,
+    )
+    return table.filter(keep)
+
+
+def _apply_template(ds: xr.Dataset, template: xr.Dataset) -> xr.Dataset:
```
alxmrs, line 72: I think this could be a public method. It's pretty useful IMO.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+    if skipped:
+        warnings.warn(
+            f"Could not re-attach non-dim coordinates from template: {skipped}",
+            stacklevel=3,
+        )
+    return out
+
+
+def _scatter_batches_to_ndarray(
+    batches: list[pa.RecordBatch],
```
alxmrs, line 151: Does this mean all the record batches are materialized into a list? I wonder if the point of the method is to put them into memory. If not, then I worry that this will break the lazy/streaming contract.

alxmrs: Reading the implementation, I see this is to put things into memory. I look forward to seeing how this is used: we should be cautious here nevertheless wrt keeping memory usage low.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+        np.full(out_shape, np.nan, dtype=dtype)
+        if np.issubdtype(dtype, np.floating)
+        else np.empty(out_shape, dtype=dtype)
+    )
+
+    # ``requested[d]`` may be in any order (callers can iselect arbitrary
+    # positions, and template coords like air_temperature.lat are descending).
+    # ``np.searchsorted`` requires ascending input, so we sort each requested
+    # array once, search there, and remap back to the original positions.
+    sorted_idx = {d: np.argsort(requested[d]) for d in dimension_columns}
```
alxmrs, line 185: This seems very reasonable. A random idea: is this something we could hoist for better performance? Probably not, since this method seems to process chunks of RecordBatches.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+    dtype: np.dtype,
+    drop_axes: list[int],
+) -> np.ndarray:
+    """Scatter Arrow ``RecordBatch`` rows into an N-D numpy buffer.
+
+    Each batch carries dim columns and a single value column for
+    ``var_name``. For every row we look up the position of its dim
+    coordinate values in the caller's requested coord arrays via
+    ``np.searchsorted`` and write the value at that N-D index. NaN-fill
+    initialization handles missing combinations (sparse results).
```
alxmrs: I'm sorry, I don't understand this docstring. Instead of saying what it does, it might be useful to say what problem it solves or explain why this method was created.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+    for batch in batches:
+        if batch.num_rows == 0:
+            continue
+        schema_names = batch.schema.names
+        # Build per-dim position arrays for this batch (positions within
+        # the caller's requested coord order).
+        positions = []
+        for d in dimension_columns:
+            col_arr = batch.column(schema_names.index(d))
+            vals = col_arr.to_numpy(zero_copy_only=False)
```
alxmrs, line 197: What are the consequences of this memory copy? Is this essential (from what I gather so far, it is).

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+        )
+        out[tuple(positions)] = value_arr.astype(dtype, copy=False)
+
+    if drop_axes:
+        out = np.squeeze(out, axis=tuple(drop_axes))
+    return cast(np.ndarray, out)
+
+
+class SQLBackendArray(xr.backends.BackendArray):
+    """Lazy N-D array backed by DataFusion's native filter pushdown.
```
alxmrs: Might be useful to state that it only supports reads today ("read only").

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+
+    On each ``__getitem__`` call, the requested xarray indexer is
+    translated into a DataFusion filter expression (``df.filter(expr)``)
+    and a column projection (``df.select(*cols)``). The filtered
+    DataFrame is consumed via ``execute_stream`` as a sequence of Arrow
+    ``RecordBatch`` es and scattered into a preallocated numpy buffer,
+    so only the requested slab is materialized. No pandas hop and no
+    SQL string synthesis.
```
alxmrs: Would be nice to explain what problem this class solves (at a high level) before describing how it solves that problem. I find it useful when reading docstrings to know how modular classes should relate to other components.

alxmrs: Please remove "no pandas hop and no SQL string synthesis".

alxmrs: Other useful things to include are the system implications of the class. You do this well so far by mentioned that only the requested slab is materialized. 

Other things that come to mind are possible warnings or errors that could be raised.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+
+    Constructed by :func:`_lazy_to_xarray`; users should not instantiate
+    this class directly.
```
alxmrs: This is useful.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+            xr.core.indexing.explicit_indexing_adapter(
+                key,
+                self.shape,
+                xr.core.indexing.IndexingSupport.OUTER,
+                self._raw_getitem,
+            ),
+        )
+
+    def __copy__(self) -> "SQLBackendArray":
+        # The backend is read-only; the underlying DataFusion DataFrame
```
alxmrs, line 283: The concern about pickling might be useful to call out in the docstring above.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+        ``key`` is a tuple of ``int``/``slice``/1-D integer-array, one per
+        dim, in :attr:`_dimension_columns` order.
+        """
+        requested: dict[str, np.ndarray] = {}
+        # Dims whose indexer covers the full extent (slice(None) or
+        # equivalent). For these we omit the filter predicate entirely
+        # so DataFusion doesn't have to evaluate a tautology.
+        full_dims: set[str] = set()
+        drop_axes: list[int] = []
+        for axis, (dim, k) in enumerate(zip(self._dimension_columns, key)):
```
alxmrs: Do we want a strict zip?

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+            coord = self._coord_arrays[dim]
+            if isinstance(k, slice):
+                start = 0 if k.start is None else k.start
+                stop = len(coord) if k.stop is None else k.stop
+                step = 1 if k.step is None else k.step
+                requested[dim] = np.asarray(coord[start:stop:step])
```
alxmrs, line 313: nice

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+            else:
+                arr = np.asarray(k)
+                requested[dim] = np.asarray(coord[arr])
+                if (
+                    len(arr) == len(coord)
+                    and (arr == np.arange(len(coord))).all()
+                ):
+                    full_dims.add(dim)
+
+        out_shape = tuple(len(requested[d]) for d in self._dimension_columns)
```
alxmrs, line 328: nice

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+            squeezed = (
+                np.squeeze(empty, axis=tuple(drop_axes)) if drop_axes else empty
+            )
+            return cast(np.ndarray, squeezed)
+
+        # Build a single DataFusion filter expression as the AND of per-dim
+        # predicates. For a single requested value: equality. For multiple:
+        # OR-chain of equalities (DataFusion 52.0.0 does not expose a clean
+        # ``Expr.in_list`` from Python; OR-chained equalities constant-fold
+        # equivalently and stay typed).
```
alxmrs, line 340: Let's file a TODO and Github issue related to this comment so we know to try to fix this when DF is upgraded.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+                for v in vals[1:]:
+                    eq = eq | (col(f'"{dim}"') == literal(v))
+                predicates.append(eq)
+
+        filtered = self._inner_df
+        if predicates:
+            combined = predicates[0]
+            for p in predicates[1:]:
+                combined = combined & p
+            filtered = filtered.filter(combined)
```
alxmrs, line 359: Nice reduce then filter.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+        batches = [b.to_pyarrow() for b in projected.execute_stream()]
+        return _scatter_batches_to_ndarray(
+            batches=batches,
+            dimension_columns=self._dimension_columns,
+            requested=requested,
+            var_name=self._var_name,
+            out_shape=out_shape,
+            dtype=self.dtype,
+            drop_axes=drop_axes,
+        )
```
alxmrs, line 376: This seems like an area to be very cautious. I want to make sure my understanding is correct: we pre-filter the data in `projected` using the indexing logic in this backend array s.t. we're guaranteed to only get a small slice, then we get that slice and convert it to an ndarray. Is that right? Or, will this pull the whole dataset into memory? Forgive me for being paranoid. 

I'm glad we're moving the batches to pyarrow RecordBatches.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+
+
+def _is_pushdownable(
```
alxmrs: Nice

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+    template: xr.Dataset | None,
+    sparse_extent: SparseExtent,
+    fill_value: Any,
+) -> xr.Dataset:
+    """Build a lazy ``xr.Dataset`` whose data vars are :class:`SQLBackendArray`.
+
+    Coord arrays are discovered per-dim via ``inner_df.select(col(d))
+    .distinct().sort(...).execute_stream()`` (one Arrow-native DataFrame
+    chain per dim). No SQL strings, no regexes; coord discovery cost is
+    bounded by N single-column scans where N is the number of dims.
```
alxmrs:
```suggestion
    chain per dim). Coord discovery cost is bounded by N single-column
    scans where N is the number of dims.
```

xarray_sql/ds.py
alxmrs: Since this is publicly facing, will you please propose several alternative names to this type? This will help me get a better understanding of the concept. Together, we can decide on a name.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+        )
+
+    schema = inner_df.schema()
+    field_names = [f.name for f in schema]
+    field_types = {f.name: f.type for f in schema}
+
+    coord_arrays: dict[str, np.ndarray] = {}
+    for d in dimension_columns:
+        dim_only = (
+            inner_df.select(col(f'"{d}"')).distinct().sort(col(f'"{d}"').sort())
```
alxmrs, line 453: What are we sorting over exactly?

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+        )
+        chunks = [b.to_pyarrow() for b in dim_only.execute_stream()]
```
alxmrs, line 455: How expensive is this operation? Does this materialize all the record batches in a big list? If it does, is this just materializing lightweight metadata?

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+        backend = SQLBackendArray(
+            inner_df=inner_df,
+            var_name=name,
+            dimension_columns=dimension_columns,
+            coord_arrays=coord_arrays,
+            shape=shape,
+            dtype=np_dtype,
+        )
+        lazy = xr.core.indexing.LazilyIndexedArray(backend)
+        data_vars[name] = xr.Variable(dimension_columns, lazy)
```
alxmrs, line 478: Nice

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+) -> xr.Dataset:
+    """Convert a tabular result to an ``xr.Dataset`` eagerly.
+
+    Used by :meth:`XarrayDataFrame.to_dataset` for queries that cannot be
+    lazily pushed down (aggregations where the result has columns not in
+    the registered template). The data path is Arrow-native: input is
+    materialized as a single ``pa.Table`` and each data variable is
+    scattered into a numpy buffer via :func:`_scatter_batches_to_ndarray`.
+    No pandas hop in the common case (the XarrayDataFrame input goes
+    straight through ``execute_stream`` to Arrow).
```
alxmrs: please remove

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+        }
+        if indexers:
+            ds = ds.reindex(indexers, fill_value=fill_value)
+
+    if template is not None:
+        ds = _apply_template(ds, template)
+    return ds
+
+
+def _eager_to_xarray(
```
alxmrs: Do we need this function?

alxmrs: If so, there seems to be repetition between this method and the above lazy version.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+    dimension_columns: list[str],
+    template: xr.Dataset | None = None,
+    sparse_extent: SparseExtent = "result",
+    fill_value: Any = np.nan,
+) -> xr.Dataset:
+    """Convert a tabular result to an ``xr.Dataset`` eagerly.
+
+    Used by :meth:`XarrayDataFrame.to_dataset` for queries that cannot be
+    lazily pushed down (aggregations where the result has columns not in
+    the registered template). The data path is Arrow-native: input is
```
alxmrs: Even if we don't have pushdowns, we still want laziness, right? We still can operate on streams.

xarray_sql/ds.py
alxmrs: I'm skeptical if we'd ever want to use this in practice.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+        self,
+        inner: Any,
+        registry: _RegistryView | None = None,
+    ) -> None:
+        object.__setattr__(self, "_inner", inner)
+        object.__setattr__(self, "_registry", registry or _RegistryView())
+
+    def to_pandas(self) -> pd.DataFrame:
+        """Materialize the result as a ``pd.DataFrame`` (DataFusion API)."""
+        return self._inner.to_pandas()
```
alxmrs, line 659: Don't we need to do this for every normal dataframe or public method? Why or why not?

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+
+    def to_dataset(
+        self,
+        dimension_columns: list[str] | None = None,
+        template: xr.Dataset | None = None,
+        template_table: str | None = None,
+        sparse_extent: SparseExtent = "result",
```
alxmrs: I almost think "sparsity" is a better name for this. Is my understanding correct? If we use "template", we get a mostly sparse xarray Dataset, but it follows the original dimensions of the cube. Whereas if we use "result", we get a dense Xarray dataset, but it is of a different, new cube representing the relation.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+                sparse_extent=sparse_extent,
+                fill_value=fill_value,
+            )
+        return _eager_to_xarray(
+            self,
+            dimension_columns=dimension_columns,
+            template=template,
+            sparse_extent=sparse_extent,
+            fill_value=fill_value,
+        )
```
alxmrs: Even if we cannot do a pushdown optimization, we will always want to use laziness. These are orthogonal, complementary concepts.

xarray_sql/ds.py
```diff
@@ -0,0 +1,800 @@
...
+        """Pick a template Dataset for metadata recovery.
+
+        Priority:
+          1. Explicit ``template_table`` argument.
+          2. If exactly one Dataset is registered on the context, use it.
+          3. None.
+
+        No SQL parsing is involved: option 2 reads only the registry's
+        contents. If multiple Datasets are registered, the caller must
+        pass ``template=`` or ``template_table=`` explicitly.
```
alxmrs: I think this is left over from the code conversation. Probably should remove.

xarray_sql/sql.py
```diff
@@ -5,12 +5,23 @@
...
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Track registered xarray Datasets so XarrayDataFrame can recover
+        # defaults (dimension_columns) and metadata (var/dataset attrs,
+        # non-dim coords, dim-coord dtype) that the forward pivot drops.
+        # Keys are the fully-qualified table names users will reference
+        # in SQL (e.g. ``"air"`` for a uniform-dim Dataset, or
+        # ``"era5.surface"`` for one entry from a multi-dim-group split).
+        self._registered_datasets: dict[str, xr.Dataset] = {}
```
alxmrs, line 23: Optional: what do you think about adding this to a post init method?

alxmrs (changes requested):
Thanks for another round of changes! Here are my thoughts. Major concerns are about laziness and naming. Last: before sending me another patch, I recommend trying this agentic development technique - adversarial iterative refinement: https://gist.github.com/dollspace-gay/45c95ebfb5a3a3bae84d8bebd662cc25

I use it myself and I find that it's good at catching issues before I make a manual self-review.

xarray_sql/ds.py
ghostiee-11: Done as `sparsity` per your other comment. Briefly considered `reindex_to_template: bool` (collapses to a bool, pairs with `fill_value`) and `extent` (drops the "sparse" word) before going with your suggestion. Type alias is `Sparsity` now.

xarray_sql/ds.py
ghostiee-11: Yes, the RecordBatches are pulled into a Python list. That list IS the scatter target. For aggregation results the size is one row per GROUP BY tuple, so it stays small. For filtered slabs it is exactly what the filter returned, no more. Worth watching if the slab grows large, but the alternative streaming-scatter would not save much without a way to skip the dense numpy buffer at the end.

xarray_sql/ds.py
ghostiee-11: The sorted positions per dim are computed once per `_raw_getitem` call and reused across every batch in that call. Hoisting across calls would not help because each call has its own `requested` arrays from a different indexer.

xarray_sql/ds.py
ghostiee-11: Yes essential. PyArrow buffers are immutable views over Arrow memory; the scatter target is a writable numpy buffer. The copy is one read pass per batch column, no temporaries.

xarray_sql/ds.py
ghostiee-11: Correct. The indexer becomes a typed predicate, `filtered = self._inner_df.filter(predicate)`, then `projected.select(*cols).execute_stream()` streams only the matching rows. We never load the full Dataset. The only time we materialize the whole result is when the user explicitly asks (a full slice or `.compute()`).

xarray_sql/ds.py
ghostiee-11: Per-dim coord column. The chain is `inner_df.select(col(d)).distinct().sort(col(d).sort())`: project one column, dedupe it, sort that column ascending. The result becomes the dim coord values in deterministic order. One small query per dim.

xarray_sql/ds.py
ghostiee-11: Just lightweight metadata. Per-dim discovery scans one column and dedupes; the list holds RecordBatches for the distinct values of that column only (kilobytes for a typical Dataset). Not the bulk of the data.

xarray_sql/ds.py
ghostiee-11: We do not need it explicitly. `__getattr__` forwards every DataFusion method we have not shadowed, so `.to_pandas()`, `.collect()`, `.count()`, `.schema()`, `.show()` all just work. I left `to_pandas` explicit for autodoc and tab-completion only. Happy to drop it if you prefer the minimum surface.

xarray_sql/sql.py
ghostiee-11: `XarrayContext` extends `SessionContext` and is not a dataclass, so `__post_init__` does not fire here. Could move the dict creation into the first `from_dataset` call lazily, but the current `__init__` body is already two lines. Happy to refactor if you have a specific pattern in mind.

ghostiee-11:
@alxmrs round 3 is up at `50b9f52`. Quick summary of what changed:

**Architecture**
- Removed the eager/lazy split. There is now a single lazy path through `SQLBackendArray`; aggregation results just produce a one-row-per-group lazy view. Pushdown and laziness are orthogonal as you said.
- Made `apply_template` public and exported it from `xarray_sql`.
- Dropped the `_RegistryView` dataclass in favor of a plain dict.

**Naming**
- `dim_cols` -> `dimension_columns`
- `SparseExtent` -> `Sparsity` (`sparsity="result"|"template"`), per your suggestion.

**Cleanup**
- Removed `_is_pushdownable`, `_eager_to_xarray`, `_to_arrow_table`, `_drop_null_dim_rows_arrow`, the bare `except Exception` fallback in `_result_columns`.
- Rewrote the `SQLBackendArray` docstring to cover purpose, read-only, pickling caveat, and errors.
- Added `strict=True` to the zip in `_raw_getitem`.
- Module docstring is now a single declarative sentence.

**Verification**
- 152/152 tests pass, ruff clean, format clean, mypy delta 0.
- End to end: sparsity result/template, aggregation+isel, full round-trip identity, public `apply_template`.
- Also did an adversarial VDD-style pass on the diff using the gist you shared, that is where the bare `except Exception` got caught. Thanks for the pointer, it is a nice discipline and I will keep using it on future PRs.

Filed #170 for the DataFusion struct-unique TODO.

Ready for re-review.

tests/test_ds.py
```diff
@@ -0,0 +1,595 @@
...
+    first_var = next(iter(source.data_vars))
+    source[first_var].attrs["units"] = "test_units"
+
+    ctx = XarrayContext()
+    ctx.from_dataset("t", source)
+    out = ctx.sql("SELECT * FROM t").to_dataset().compute()
+
+    sort_keys = list(out.dims)
+    actual = _clear_encoding(out.sortby(sort_keys))
+    expected = _clear_encoding(source.compute().sortby(sort_keys))
```
alxmrs, line 118: Is `sortby` necessary? Maybe we can file a follow up issue to this one to make the round-trips "stable" -- that is, preserving of dimension order.

tests/test_ds.py
```diff
@@ -0,0 +1,595 @@
...
+            template_table="air",
+        )
+
+
+# ---------------------------------------------------------------------------
+# Lazy backend semantics
+# ---------------------------------------------------------------------------
+
+
+def test_lazy_select_star_returns_lazily_indexed_array(air_dataset_small):
```
alxmrs: 🐑 I could go either way on keeping this test. It seems to verify the implementation detail rather than the interface. Though, tbd, I like the test.

tests/test_ds.py
```diff
@@ -0,0 +1,595 @@
...
+    # xr.core.indexing.LazilyIndexedArray wraps our SQLBackendArray.
+    assert "LazilyIndexedArray" in type(inner).__name__
+    from xarray_sql.ds import SQLBackendArray
+
+    # Drill in to confirm the underlying array is ours.
+    underlying = inner.array if hasattr(inner, "array") else inner
+    assert isinstance(underlying, SQLBackendArray)
+
+
+def test_lazy_no_slab_access_until_data_read(air_dataset_small, monkeypatch):
```
alxmrs: I'm not sure what a "slab" is, but I like this test.

tests/test_ds.py
```diff
@@ -0,0 +1,595 @@
...
+    """Reading metadata (``dims``, ``coords``, ``attrs``) on the lazy
+    Dataset must not trigger ``SQLBackendArray._raw_getitem`` (the slab
+    materialization entry point)."""
+    from xarray_sql import ds as ds_mod
+
+    ctx = XarrayContext()
+    ctx.from_dataset("air", air_dataset_small)
+    out = ctx.sql("SELECT * FROM air").to_dataset()
+    calls = {"n": 0}
+    original = ds_mod.SQLBackendArray._raw_getitem
```
alxmrs: 🐑 optional: it might be cleaner to create a fake BackendArray that is a subclass of `SQLBackendArray` for the monkeypatch (then in the method, we could just call `super()`). This is good though.

tests/test_ds.py
```diff
@@ -0,0 +1,595 @@
...
+        air_dataset_small["air"]
+        .compute()
+        .isel(time=slice(0, 3))
+        .sortby(["lat", "lon"])
+        .values
+    )
+    np.testing.assert_array_equal(actual, expected)
+
+
+def test_lazy_select_star_round_trip_equality(air_dataset_small):
```
alxmrs: Don't we have another round trip test above?

alxmrs: Ideally, we could have a really simple round trip test, but run it on a varied number of dataset inputs (e.g. we could have a fixture that returned all the xarray tutorial datasets).

tests/test_ds.py
```diff
@@ -0,0 +1,595 @@
...
+    """Lazy .values produces the same data as the eager path on full read."""
+    ctx = XarrayContext()
+    ctx.from_dataset("air", air_dataset_small)
+    out = ctx.sql("SELECT * FROM air").to_dataset()
+    expected = air_dataset_small.compute().sortby(["time", "lat", "lon"])
+    actual = out.sortby(["time", "lat", "lon"])
+    np.testing.assert_array_equal(actual["air"].values, expected["air"].values)
+
+
+def test_aggregation_uses_lazy_backend(air_dataset_small):
```
alxmrs: To me, this tests the implementation detail, not the contract. A better test for this capability would be to chunk the input dataset in such a way and verify that it is loaded in a "streaming" method.

tests/test_ds.py
```diff
@@ -0,0 +1,595 @@
+"""Tests for the SQL -> xarray reverse path.
```
alxmrs, line 1: Let's prompt Claude to review these tests and pair them down to testing essential functionality. Having too many tests puts a high maintenance burden on the project (tests require human verification) to the point that many of them might not be worth the regressions they are designed to catch. How can we pair down tests to cover orthogonal items of the contract? How can tests cover important edge cases, especially those that guard against areas of the code that are likely to change?

In general, let's favor "integration style" tests that verify multiple connection points of components, especially ones that verify a core contract that we hope to deliver. When possible, let's write "simulation style" tests that create an abstract, verifiably contract (like a round trip should return the input) and vet it using a wide range of data (even randomly generated datasets).

alxmrs: Thinking a bit more: can we cover the same test cases provided by the tests written here today in a more succinct way?

xarray_sql/__init__.py
```diff
@@ -1,11 +1,14 @@
...
+from .ds import XarrayDataFrame, apply_template
 from .reader import read_xarray, read_xarray_table
 from .sql import XarrayContext
-from .df import from_map
 
 __all__ = [
     "cftime",
     "XarrayContext",
+    "XarrayDataFrame",
+    "apply_template",
```
alxmrs: @ahuang11 I'm torn here, and I could use a second opinion: Should these two components be exposed in the public API for this package? My current mind says that these should be internal implementation details.

I think a better public function to expose (and to write here now or later) would be a `to_dataset()` function that took `XarrayDataFrame` and produced an `xarray.Dataset`.

xarray_sql/ds.py
```diff
@@ -0,0 +1,622 @@
...
+import xarray as xr
+from datafusion import col, literal
+
+# ``sparsity`` selects the dim-coord range of the output for a
+# filtered query:
+#   - "result"   : keep only the dim values present in the query result
+#                  (sparse output equal to whatever rows came back).
+#   - "template" : reindex to the registered Dataset's full coord ranges,
+#                  filling absent cells with ``fill_value``.
+Sparsity = Literal["result", "template"]
```
alxmrs: Thanks for this change. Optional nit: Maybe this comment could be a Python docstring.

xarray_sql/ds.py
alxmrs: This can happen in a follow up issue, but I'd love to see if your agent could write a performance minded test on realistic data to see if materialized data always remained small or if it commonly exploded memory. So far, the tutorial datasets and other datasets are toy and thus not a faithful way to verify if this is a serious concern.

alxmrs: Also: what is a "slab"?

xarray_sql/ds.py
```diff
@@ -0,0 +1,622 @@
...
+def _scatter_batches_to_ndarray(
+    batches: list[pa.RecordBatch],
+    dimension_columns: list[str],
+    requested: dict[str, np.ndarray],
+    var_name: str,
+    out_shape: tuple[int, ...],
+    dtype: np.dtype,
+    drop_axes: list[int],
+) -> np.ndarray:
+    """Convert filtered Arrow ``RecordBatch`` rows into a dense N-D numpy slab.
```
alxmrs: I am not familiar with "slab" -- can we use a term of art here, instead?

xarray_sql/ds.py
alxmrs: Thanks!

xarray_sql/ds.py
alxmrs: Experimental idea, could happen in a follow up: the contract for `xarray.Variable` is not a numpy array, but anything that is numpy-array-like: https://github.com/pydata/xarray/blob/a1408234d1dc786965bfb6311f8782561ee0e676/xarray/core/variable.py#L373

Given this, I suspect we could find an arrow interface that also meets the numpy array contract. At a high level, we'd just point to the underlying array buffers in the record batches without copies directly from Xarray. That would be much more performant if we could pull it off.

alxmrs (commented):
First half of review feedback on this round. I'll come back later today if I have time, more likely tomorrow.

ghostiee-11:
@alxmrs round 4 is up at `7b7518b`. Short notes:

- Made `XarrayDataFrame` and `apply_template` internal. `apply_template` is now `_apply_template`. Only `XarrayContext` is exported, and `ctx.sql().to_dataset()` stays the user-facing path. If you and @ahuang11 land on a top-level `to_dataset(df)` function later, easy to add then.
- Dropped "slab" everywhere. Used plain language ("the requested data", "the indexed region") in docstrings and renamed `slab = ...` in the README example to `region`.
- Converted the `sparsity` inline comment to a proper attribute docstring on the type alias.
- Paired down tests 31 to 28. Dropped the call-count / isinstance / duplicate ones (`test_lazy_no_slab_access_until_data_read`, `test_full_dim_slice_omits_filter_for_full_dims`, `test_lazy_select_star_returns_lazily_indexed_array`, `test_aggregation_uses_lazy_backend`, `test_lazy_select_star_round_trip_equality`). Added `eraint_uvz` to the parametrized round-trip identity test so we cover one xarray tutorial dataset on top of the conftest fixtures.

148/148 pass, ruff clean, format clean. Net -140 lines.

Will file the three follow-ups you mentioned (stable round-trip / dim order, perf test on realistic data, zero-copy Arrow buffer view) so we do not lose them.

alxmrs:
I want to get this merged, it's a significant, positive contribution. My last step is for me to manually validate the changes on my local machine. I'm traveling today, but I'd do this as time allows.

alxmrs:
Maybe it's my location, but when I do the inverse operation on the ERA5 dataset from the README example, I notice that the `to_dataset()` call is really slow. Here's what I'm doing: 
```
>>> result = ctx.sql('''
...   SELECT level, AVG(temperature) - 273.15 AS avg_c
...   FROM era5.atmosphere
...   WHERE time BETWEEN TIMESTAMP '2020-01-01'
...                  AND TIMESTAMP '2020-01-01 05:00:00'
...   GROUP BY level
...   ORDER BY level DESC
... ''')
>>> result
DataFrame()
+-------+----------------------+
| level | avg_c                |
+-------+----------------------+
| 1000  | 6.6210120796502565   |
| 975   | 5.185637919348153    |
| 950   | 4.028428657263021    |
| 925   | 3.0828117974912743   |
| 900   | 2.2109172992531967   |
| 875   | 1.395017610194202    |
| 850   | 0.6342670572626616   |
| 825   | -0.21037158786759846 |
| 800   | -1.1810754318269687  |
| 775   | -2.3064649711534457  |
+-------+----------------------+
Data truncated.
>>> agg_ds = result.to_dataset(dimension_columns=["level"]) # this takes a surprisingly long time
```
Any idea why this could be so slow? I would hope that it would return immediately with an overview of the dataset, and only materialize chunks in a lazy manner. 

Also, now that I see how this is used in context (I needed to try it out) I think "dimension_columns" should be renamed to just "dims".

ghostiee-11:
@alxmrs thanks for jumping in and pushing the perf fix, the chunks-based partitioning, the `dims` rename, and the unified `template` arg directly. The chunks knob is a much cleaner contract than the introspection path I had, and the ORDER BY plumbing was a nice catch.

Just pushed `1a29fa4` with a `ruff format` over the two files; that was what the 3.10 and 3.13 lint jobs were failing on. CI should be green now.

Anything else you want me to clean up before this lands?

ghostiee-11:
Quick validation of your perf fix on a 115 MB ERA5-shaped synthetic (48 hourly time chunks x 37 levels x 90 x 180), same shape of query as your repro:

| stage (cold) | pre-fix `f5f0160` | post-fix `1a29fa4` |
|---|---:|---:|
| `ctx.sql(...)` | 374 ms | 45 ms |
| `result.to_dataset(dims=["level"])` | **684 ms** | **192 ms** |
| `agg_ds["avg_c"].values` | 88 ms | 0 ms |
| total cold | 1146 ms | **237 ms (4.8x faster)** |

`.values` is free on the post-fix because the materialize happens once during construction. On real ERA5 from GCS the gap should be much larger since each redundant scan is a remote round trip.

Also confirmed:
- `ORDER BY level DESC` carries through, `out["level"].values` is strictly descending.
- `avg_c` matches `ds.mean(dim="time") - 273.15` within float32 noise (rtol=1e-4).
- 159/159 tests pass locally.

ghostiee-11:
Thanks a lot @alxmrs, for the attribution :)

PR #169, opened 2026-06-04
Support heterogeneous datasets with scalar variables in from_dataset

### LINKED ISSUE #168: Error on reading arraylake derived zarr store
Read this https://www.earthmover.io/blog/virtual-zarr looked cool, and was testing
```
import arraylake as al
import xarray as xr
import xarray_sql as xql

client = al.Client()
client.login()

repo = client.get_repo("earthmover-public/goes-16")
session = repo.readonly_session("main")
ds = xr.open_zarr(session.store, group="ABI-L2-MCMIPF/post-2023-04-19")
xql.read_xarray(ds)
```

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[9], line 11
      7 
      8 repo = client.get_repo("earthmover-public/goes-16")
      9 session = repo.readonly_session("main")
     10 ds = xr.open_zarr(session.store, group="ABI-L2-MCMIPF/post-2023-04-19")
---> 11 xql.read_xarray(ds)

File ~/miniconda3/envs/lumen/lib/python3.12/site-packages/xarray_sql/reader.py:184, in read_xarray(ds, chunks)
    171 def read_xarray(ds: xr.Dataset, chunks: Chunks = None) -> pa.RecordBatchReader:
    172     """Pivots an Xarray Dataset into a PyArrow Table, partitioned by chunks.
    173 
    174     Args:
   (...)    182       Dataset.
    183     """
--> 184     reader = XarrayRecordBatchReader(ds, chunks=chunks)
    185     return pa.RecordBatchReader.from_stream(reader)

File ~/miniconda3/envs/lumen/lib/python3.12/site-packages/xarray_sql/reader.py:93, in XarrayRecordBatchReader.__init__(self, ds, chunks, batch_size, _iteration_callback)
     91 fst = next(iter(ds.values())).dims
     92 if not all(da.dims == fst for da in ds.values()):
---> 93     raise ValueError(
     94         "All dimensions must be equal. Please filter data_vars in the Dataset."
     95     )

ValueError: All dimensions must be equal. Please filter data_vars in the Dataset.```

### PR DESCRIPTION
**Pull request by ghostiee-11**
`XarrayContext.from_dataset` crashed on datasets that mix dimensions with scalar (0-dim) variables, e.g. GOES-16 imagery where `(y, x)` bands sit beside scalar metadata like `goes_imager_projection` (#168). Scalars land in a `()` group that `block_slices` couldn't partition. Now `block_slices` treats a dimensionless dataset as a single block, and `from_dataset` registers that group as a one-row `scalar` table (overridable via `table_names[()]`). Fixes #168.

**Before**  real GOES-16 file (124 vars, 89 of them scalar):
```python
>>> ctx.from_dataset("goes", ds)
AssertionError: Dataset `ds` must be chunked or `chunks` must be provided.
```

**After:**
```python
>>> ctx.from_dataset("goes", ds)
>>> sorted(ctx.catalog().schema("goes").table_names())
['number_of_image_bounds', 'number_of_time_bounds', 'scalar', 'y_x']
>>> ctx.sql("SELECT COUNT(*) AS n FROM goes.y_x").to_pydict()
{'n': [250000]}
>>> ctx.sql("SELECT * FROM goes.scalar").to_pandas().shape
(1, 89)
```

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -95,17 +95,26 @@ scalar variables such as `goes_imager_projection`. `from_dataset` groups all the
 scalars into a single one-row table named `scalar`:
 
 ```python
+import urllib.request
+
 import xarray as xr
 from xarray_sql import XarrayContext
 
-# A GOES-16 ABI cloud-and-moisture file: (y, x) bands + scalar metadata.
-ds = xr.open_dataset('OR_ABI-L2-MCMIPM1-M6_G16_....nc').chunk({'y': 250, 'x': 250})
+# A real GOES-16 ABI cloud-and-moisture file from NOAA's public bucket:
+# (y, x) image bands alongside dozens of scalar metadata variables.
+url = (
+    'https://noaa-goes16.s3.amazonaws.com/ABI-L2-MCMIPM/2024/001/00/'
+    'OR_ABI-L2-MCMIPM1-M6_G16_s20240010000281_e20240010000350_c20240010000426.nc'
+)
+urllib.request.urlretrieve(url, 'goes.nc')
+ds = xr.open_dataset('goes.nc').chunk({'y': 250, 'x': 250})
 
 ctx = XarrayContext()
 ctx.from_dataset('goes', ds)
 
-ctx.sql('SELECT COUNT(*) FROM goes.y_x')   # the gridded bands
-ctx.sql('SELECT * FROM goes.scalar')       # one row of metadata
+# The gridded bands and the scalar metadata are separate tables.
+ctx.sql('SELECT COUNT(*) AS n FROM goes.y_x').to_pandas()['n'][0]  # -> 250000
+ctx.sql('SELECT * FROM goes.scalar').to_pandas().shape            # -> (1, 89)
 ```
 
 Override the default name like any other group with `table_names={(): 'metadata'}`.
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -87,12 +87,30 @@ If you omit `table_names`, each table is named by joining its dimension names
 with underscores, e.g. `era5.time_latitude_longitude` and
 `era5.time_level_latitude_longitude`.
 
-Scalar (0-dimensional) variables, common as metadata in real-world stores such
-as GOES satellite imagery (`goes_imager_projection`, etc.), are grouped together
-into a single one-row table named `scalar` by default. Override it like any
-other group with `table_names={(): 'metadata'}`.
+## GOES satellite imagery (scalar variables)
 
-A runnable version of this example lives at
+Real-world stores often mix gridded data with scalar (0-dimensional) metadata.
+GOES satellite imagery, for example, pairs `(y, x)` image bands with dozens of
+scalar variables such as `goes_imager_projection`. `from_dataset` groups all the
+scalars into a single one-row table named `scalar`:
+
+```python
+import xarray as xr
+from xarray_sql import XarrayContext
+
+# A GOES-16 ABI cloud-and-moisture file: (y, x) bands + scalar metadata.
+ds = xr.open_dataset('OR_ABI-L2-MCMIPM1-M6_G16_....nc').chunk({'y': 250, 'x': 250})
+
+ctx = XarrayContext()
+ctx.from_dataset('goes', ds)
+
+ctx.sql('SELECT COUNT(*) FROM goes.y_x')   # the gridded bands
+ctx.sql('SELECT * FROM goes.scalar')       # one row of metadata
+```
+
+Override the default name like any other group with `table_names={(): 'metadata'}`.
+
+A runnable version of the ERA5 example lives at
 [`perf_tests/era5_temp_profile.py`](../perf_tests/era5_temp_profile.py).
 
 [arco-era5]: https://github.com/google-research/arco-era5
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -95,8 +95,7 @@ scalar variables such as `goes_imager_projection`. `from_dataset` groups all the
 scalars into a single one-row table named `scalar`:
 
 ```python
-import urllib.request
-
+import fsspec
 import xarray as xr
 from xarray_sql import XarrayContext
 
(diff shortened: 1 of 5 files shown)


docs/examples.md
```diff
@@ -87,6 +87,11 @@ If you omit `table_names`, each table is named by joining its dimension names
 with underscores, e.g. `era5.time_latitude_longitude` and
 `era5.time_level_latitude_longitude`.
 
+Scalar (0-dimensional) variables, common as metadata in real-world stores such
```
alxmrs: Since this is the examples file, let's also show an example of the GOES dataset here.

xarray_sql/sql.py
```diff
@@ -86,10 +86,12 @@ def from_dataset(
         self.catalog().register_schema(name, schema)
 
         for dims, var_names in groups.items():
-            sub_name = table_names.get(dims, "_".join(dims))
-            sub_ds = input_table[var_names]
-            schema.register_table(sub_name, read_xarray_table(sub_ds, chunks))
-            self._maybe_register_cftime_udf(sub_ds)
```
alxmrs, line 92: We lost this registration function -- don't we need this?

xarray_sql/sql.py
```diff
@@ -101,11 +103,21 @@ def _from_dataset(
     ):
         """Register a uniform-dimension Dataset as a single SQL table."""
 
-        table = read_xarray_table(input_table, chunks)
-        self.register_table(table_name, table)
-        self._maybe_register_cftime_udf(input_table)
```
alxmrs, line 106: Same here, why was this deleted?

xarray_sql/sql.py
```diff
@@ -101,11 +103,21 @@ def _from_dataset(
...
+        self._register_table(
+            self.register_table, table_name, input_table, chunks
+        )
         return self
 
+    def _register_table(self, register, table_name, ds, chunks):
```
alxmrs: Do we need this new method, or could we merely call `_from_dataset()` in the public `from_dataset()` method?

xarray_sql/sql.py
```diff
@@ -101,11 +103,21 @@ def _from_dataset(
...
+        """Register one (sub)dataset as a table via ``register``.
+
+        ``register`` is the registration callable for the target namespace:
+        ``self.register_table`` for a top-level table, or
+        ``schema.register_table`` for a table inside a SQL schema.
+        """
+        register(table_name, read_xarray_table(ds, chunks))
```
alxmrs: Where is this `register` function defined?

alxmrs (changes requested):
I like the general direction of the fix, thanks. I've raised a few points of feedback. 

Thanks for the quick turnaround on the fix.

docs/examples.md
ghostiee-11: True, added a GOES example with the scalar table.

xarray_sql/sql.py
ghostiee-11: Not lost, just moved, and I've reworked it per your comment below, so registration lives back in `_from_dataset`

xarray_sql/sql.py
ghostiee-11: Same thing, `_from_dataset` still does the registration, the rework makes that obvious now.

xarray_sql/sql.py
ghostiee-11: Agreed, done the loop calls `_from_dataset` directly now, with an optional `schema` target

xarray_sql/sql.py
ghostiee-11: It was a passed-in callable. Removed  `_from_dataset` picks self vs schema itself.

ghostiee-11:
Hey @alxmrs, request you to review this.

docs/examples.md
```diff
@@ -87,7 +87,30 @@ If you omit `table_names`, each table is named by joining its dimension names
...
+GOES satellite imagery, for example, pairs `(y, x)` image bands with dozens of
+scalar variables such as `goes_imager_projection`. `from_dataset` groups all the
+scalars into a single one-row table named `scalar`:
+
+```python
+import xarray as xr
+from xarray_sql import XarrayContext
+
+# A GOES-16 ABI cloud-and-moisture file: (y, x) bands + scalar metadata.
+ds = xr.open_dataset('OR_ABI-L2-MCMIPM1-M6_G16_....nc').chunk({'y': 250, 'x': 250})
```
alxmrs: Let's not abbreviate this path. I think the code should be copy and pastable and work.

docs/examples.md
```diff
@@ -87,7 +87,30 @@ If you omit `table_names`, each table is named by joining its dimension names
...
+
+ctx = XarrayContext()
+ctx.from_dataset('goes', ds)
+
+ctx.sql('SELECT COUNT(*) FROM goes.y_x')   # the gridded bands
```
alxmrs: Can you add python comments to show that this prints? Same with the line below.

docs/examples.md
```diff
@@ -87,7 +87,30 @@ If you omit `table_names`, each table is named by joining its dimension names
...
+ctx.sql('SELECT * FROM goes.scalar')       # one row of metadata
+```
+
+Override the default name like any other group with `table_names={(): 'metadata'}`.
```
alxmrs, line 120: Thanks, this is helpful.

alxmrs (approved):
A few notes, then I think this is good to merge. Thanks for the fix! :)

docs/examples.md
ghostiee-11: Done, it now pulls a real GOES-16 file from NOAA's public bucket and runs end to end.

docs/examples.md
ghostiee-11: Added # -> output comments on both queries (250000 and (1, 89)).

ghostiee-11:
> A few notes, then I think this is good to merge. Thanks for the fix! :)

It's done.. @alxmrs I guess it's good to merge now!!

docs/examples.md
```diff
@@ -87,7 +87,39 @@ If you omit `table_names`, each table is named by joining its dimension names
...
+import xarray as xr
+from xarray_sql import XarrayContext
+
+# A real GOES-16 ABI cloud-and-moisture file from NOAA's public bucket:
+# (y, x) image bands alongside dozens of scalar metadata variables.
+url = (
+    'https://noaa-goes16.s3.amazonaws.com/ABI-L2-MCMIPM/2024/001/00/'
+    'OR_ABI-L2-MCMIPM1-M6_G16_s20240010000281_e20240010000350_c20240010000426.nc'
+)
+urllib.request.urlretrieve(url, 'goes.nc')
```
alxmrs: I don't think you need to download the netcdf file first. You should be able to open it just with Xarray, maybe first using fsspec.

alxmrs: Just your luck -- it looks like this dataset was just virtualized: https://earthmover.io/blog/virtual-zarr/

ghostiee-11: Good call. Switched to fsspec with simplecache so xarray opens it directly, no manual download step.

ghostiee-11: Ha, that is the same blog that kicked this whole thread off. I kept the NOAA file here so the example stays auth free and copy pasteable, but the virtual zarr version is really cool.

ghostiee-11:
@alxmrs all notes addressed, should be good to merge now. Thanks for the reviews!

PR #175, opened 2026-06-15
fix: coalesce native chunks into bounded scan partitions (#174)

### LINKED ISSUE #174: read_xarray registration is O(num_chunks): one partition per native chunk, intractable on finely-chunked stores
`read_xarray_table` creates one scan partition per native chunk and enumerates them all at registration, so registration cost is `O(num_chunks)` rather than `O(data_size)`. On finely-chunked stores this makes registration intractable.

### Measured (tiny synthetic data, same data, only the chunking changes)

```python
import numpy as np, xarray as xr, time, xarray_sql as xql
for T in (1_000, 10_000, 100_000):
    ds = xr.Dataset(
        {"v": (("t", "y", "x"), np.zeros((T, 2, 2), "float32"))},
        coords={"t": np.arange(T), "y": np.arange(2), "x": np.arange(2)},
    ).chunk({"t": 1, "y": 2, "x": 2})
    t0 = time.time(); xql.read_xarray_table(ds, chunks={"t": 1})
    print(T, f"{time.time()-t0:.2f}s")
```

| partitions | register time |
| --- | --- |
| 1,000 | 0.06s |
| 10,000 | 0.24s |
| 100,000 | 2.45s |

~25us/partition, linear.

### Real-world impact

Public Arraylake GOES (`earthmover-public/goes-16`, group `ABI-L2-MCMIPF/post-2023-04-19`) is chunked one timestep per chunk:

```
CMI_C13 chunks -> t: 102,988 chunks of size 1 ; y, x: 24 chunks of 226
```

That is `102988 * 24 * 24 = ~59M` partitions for a **single** variable, i.e. ~25-30 minutes just to register one variable. Registering all 188 variables never finishes. Coarsening the time chunks and pre-loading coordinates did not help (still minutes).

### Proposal

Decouple partition count from native chunk count: coalesce native chunks into a bounded number of scan partitions (e.g. a target partition count / size), or enumerate partitions lazily, so registration cost is independent of how finely the store is chunked.

### Context

Surfaced via Lumen's `XArraySQLSource` (holoviz/lumen#1889), which compounds it by registering per-variable.

ghostiee-11:
## What this does

`read_xarray_table` creates one DataFusion scan partition per native xarray chunk, so table registration is O(num_chunks) rather than O(data_size). On finely chunked stores this becomes intractable. The case in #174 is a single Arraylake GOES-16 variable with roughly 102,988 x 24 x 24 (about 59M) native chunks, which takes 25 to 30 minutes to register and never finishes for all 188 variables.

This coalesces consecutive native chunks into at most `target_partitions` scan partitions (default 16384), allocated in a balanced way across dimensions so filter pushdown keeps pruning in every dimension. Each coalesced partition still streams one native chunk at a time, so peak per-partition memory during a query is unchanged. Passing `target_partitions=None` restores the previous behavior of one partition per native chunk.

Fixes #174.

## Why this is safe

Coalescing only ever widens a partition's min/max bounding box. The pruning logic excludes a partition only when a filter provably cannot match its box, so a wider box can never cause a false exclusion. Query results are therefore identical with or without coalescing; only pruning granularity changes. An oracle test asserts this by comparing results at the default target against `target_partitions=None`.

The group-size allocation is O(D log D) in the number of dimensions and independent of the native chunk count, so it stays fast even at tens of millions of chunks.

## Results

Registration time vs native chunk count (synthetic, `perf_tests/registration_scaling.py`):

| native chunks | before (one per chunk) | after (coalesced) |
| --- | --- | --- |
| 1,000 | 0.02s | 0.05s |
| 10,000 | 0.25s | 0.25s |
| 100,000 | 2.93s | 0.32s |
| 1,000,000 | 54.27s | 0.46s |

Below the target the behavior is unchanged (no coalescing); above it, registration stays flat.

## API

- `read_xarray_table(ds, ..., target_partitions=16384)`
- `XarrayContext.from_dataset(..., target_partitions=16384)` (the bound is applied per table)
- `target_partitions=None` disables coalescing.

## Testing

- `tests/test_coalesce.py`: group-size algorithm, exact sub-block tiling, bounded partition count, correctness oracle (coalesced vs uncoalesced including WHERE filters), bounded-memory streaming, explicit-chunks capping, and cftime (360_day) transparency.
- `perf_tests/registration_scaling.py`: registration-scaling benchmark.
- Full suite passes: `uv run --no-project pytest -v . -m "not integration"` (155 tests). `cargo check --all-features` passes. No Rust changes.
- Docstrings updated (Google style) and a note added to the docs.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -35,6 +35,11 @@ Open a year of ARCO-ERA5 and let SQL `WHERE` clauses do the filtering — the
 library prunes time partitions and pushes dimension-column filters down. Use
 the `table_names` kwarg to give each dimension group a friendly name:
 
+Native chunks are coalesced into at most `target_partitions` scan partitions
+(default 16384), so registration stays fast even on stores with millions of
+fine chunks. Raise `target_partitions` for more selective pruning, or pass
+`None` to keep one partition per native chunk.
+
 ```python
 import xarray as xr
 import xarray_sql as xql
--- a/perf_tests/registration_scaling.py
+++ b/perf_tests/registration_scaling.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+"""Registration time vs. native chunk count (issue #174).
+
+`read_xarray_table` used to create one scan partition per native xarray chunk,
+making registration O(num_chunks): ~25 us/partition, so a finely chunked store
+(e.g. a GOES-16 variable with ~59M native chunks) took tens of minutes just to
+register, if it finished at all.
+
+Native chunks are now coalesced into at most `target_partitions` scan
+partitions, so registration cost is bounded regardless of how finely the store
+is chunked. This script registers a synthetic dataset at a range of native
+chunk counts and prints registration time with coalescing on (bounded) vs. off
+(`target_partitions=None`, the historical O(num_chunks) behavior).
+
+No network or large memory needed: the dataset is tiny per-chunk; only the
+*number* of chunks grows, which is exactly what drives registration cost.
+
+Run: python perf_tests/registration_scaling.py
+"""
+
+import time
+
+import numpy as np
+import pandas as pd
+import xarray as xr
+
+import xarray_sql as xql
+from xarray_sql.df import DEFAULT_TARGET_PARTITIONS
+
+
+def make_dataset(n_time_chunks: int) -> xr.Dataset:
+    """A (time, lat, lon) dataset with `n_time_chunks` native time chunks."""
+    time_coord = pd.date_range("2000-01-01", periods=n_time_chunks, freq="h")
+    lat = np.linspace(-90, 90, 4)
+    lon = np.linspace(-180, 180, 4)
+    data = np.zeros((n_time_chunks, 4, 4), dtype="float32")
+    ds = xr.Dataset(
+        {"v": (["time", "lat", "lon"], data)},
+        coords={"time": time_coord, "lat": lat, "lon": lon},
+    )
+    # One native chunk per time step -> n_time_chunks native partitions.
+    return ds.chunk({"time": 1, "lat": 4, "lon": 4})
+
+
+def time_registration(ds: xr.Dataset, target_partitions) -> float:
+    t0 = time.perf_counter()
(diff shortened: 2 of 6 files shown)


ghostiee-11:
Hey @alxmrs, can you review this?

alxmrs:
After merging my speedup changes with this PR, I found this turned ERA5 registration from 10s --> 1.2s. But, it made a query that previously took 1.4s to take 24s. Would you mind exploring more to see if we can get both the win of faster registration and fast queries? Please see if you can compare to main again -- esp, on the GOES example -- because I may have independently fixed the issue using a different means (addressing dask).

ghostiee-11:
Thanks for testing this against dask @alxmrs . I looked into both sides before changing anything.

The query regression is real: coalescing prunes to a whole coalesced group instead of a single chunk, so a selective filter reads many times more than it needs (the 1.4s to 24s you saw).

But coalescing is the only thing making GOES-scale registration tractable. On current main, registration is still O(num_chunks): 1M chunks take ~15s, so one 59M-chunk GOES variable is ~15 min, and there are 188. So #174 is still open at that scale, and this PR fixes it the wrong way, by sacrificing pruning.

I think we can get both, building on #176. The pruning metadata is separable per dimension: a partition's min/max box is fixed by per-axis chunk bounds, not the full product (about 103k per-axis values vs 59M boxes for GOES). So:

1. Store per-axis chunk metadata at registration. Bounded, O(sum of per-axis chunk counts).
2. Prune per axis at scan time and enumerate only the surviving partitions, at full native granularity. No pruning lost.
3. Coalesce only the survivors of an unfiltered scan, purely for parallelism.

That gives bounded registration and exact pruning. The cost is reworking the Rust table provider to prune-then-enumerate instead of materialising all N partitions up front.

Want me to prototype this so we can benchmark it on the GOES example? If you would rather keep it small for now, I can default `target_partitions` to `None` so this PR is opt-in and changes nothing by default.

alxmrs:
Can you paste in a comment or the change a demo script of opening GOES? I'd like to make a few sanity checks to see if there are other ways we could make it open faster. For example, do you set `chunks=None` when you open the Zarr?

PR #177, opened 2026-06-23
Geospatial SQL benchmark suite: prove core geo/climate ops are relational

alxmrs:
## What & why

A suite of worked examples in `benchmarks/geospatial/` arguing a deliberately radical thesis: **the core operations of geospatial and climate analysis — the ones we reach for an array library to perform — are, underneath, relational operations.** Climatologies, anomalies, zonal means, spectral indices, forecast skill, even reprojection and regridding map onto ordinary SQL (`GROUP BY`, `JOIN`, window functions, `CASE`, and the occasional scalar UDF).

Every case poses the operation in SQL against `xarray-sql`, **round-trips the result back to an `xarray.Dataset`/`DataArray`** (via `to_dataset`), and **asserts it matches an xarray reference** (coordinate-aligned `xr.testing.assert_allclose`). SQL in, an array out, verified — and each run prints the result's standard xarray repr. The headline is correctness + clarity of the SQL, not raw speed.

The README gains a **"Does it work?"** section summarizing the findings (addresses #46), and a full write-up lives in [`docs/geospatial.md`](https://github.com/alxmrs/xarray-sql/blob/geospatial-sql-benchmarks/docs/geospatial.md) (wired into the docs nav).

## The cases (all validated end-to-end on real data)

| # | Operation | Relational form | Data + reference |
|---|-----------|-----------------|------------------|
| 01 | NDVI | column arithmetic (nodata auto-masked on decode) | real **Sentinel-2 L2A** Zarr (ESA EOPF, via `pystac-client` + `xr.open_datatree`) |
| 02 | Diurnal climatology | `GROUP BY lat, lon, hour` | ARCO-ERA5 |
| 03 | Zonal mean | `GROUP BY latitude` | full ARCO-ERA5, `WHERE`-pruned |
| 04 | Anomaly | climatology CTE self-`JOIN` | ARCO-ERA5 |
| 05 | **Forecast skill** | forecast↔truth `JOIN` on `valid_time = init + lead` | **WeatherBench 2 Pangu & GraphCast vs ERA5** |
| 06 | Zonal stats over regions | raster × vector range `JOIN` | full ARCO-ERA5 + continental boxes |
| 07 | Reprojection | PROJ scalar **UDF** (`ST_Transform`-style) | **Earth Engine via Xee**, validated against EE's own `pixelLonLat` |
| 08 | Regridding | sparse weight-table `JOIN` + weighted `GROUP BY` | real **SRTM elevation** via Xee, vs xarray `.interp()` |

[...]

PR #178, opened 2026-06-24
Add lazy round-trip benchmark (case 09)

ghostiee-11:
For #177 , Adds `09_lazy_roundtrip.py`: the same `air[t0]` slab is 1,325 rows / ~0.7 MB via lazy `to_dataset(chunks={"time":1})` + `.sel(time=t0)` (one WHERE pushed down) vs 3.86M rows / ~161 MB eager, all asserted equal to the xarray reference.
Also hardens `_harness.py` (grid-coverage guard in `assert_grid_close`, tracemalloc `finally` in `measured`); runs green locally via uv.

### DIFF
--- a/benchmarks/geospatial/09_lazy_roundtrip.py
+++ b/benchmarks/geospatial/09_lazy_roundtrip.py
@@ -1,0 +1,143 @@
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#   "xarray-sql",
+#   "xarray",
+#   "numpy",
+#   "pandas",
+#   "dask",
+#   "pooch",
+#   "netCDF4",
+# ]
+#
+# [tool.uv.sources]
+# xarray-sql = { path = "../../", editable = true }
+# ///
+"""Lazy round-trip: the SQL answer comes back as an array without materializing.
+
+The other cases prove the SQL computes the *same numbers* as xarray. This one
+proves the other half of the claim the suite leans on: the round-trip back to
+xarray is **lazy**. ``ctx.sql(...).to_dataset()`` hands you a Dataset whose data
+is still a query; slicing it (``.sel(time=t0)``) pushes a ``WHERE`` back down
+into SQL, so reading one slab reads one slab, not the whole table.
+
+That is the property the Large Scale Geospatial Benchmarks discussion
+(coiled/benchmarks #1545) actually asks about: not "can you express it" but
+"does the stack stay light when you point it at a big archive and pull a slice".
+Here we make it a number. Three ways to get one timestep out of SQL:
+
+    eager   ctx.sql(...).to_pandas()                       # whole long table
+    eager   to_dataset(chunks=None)[v].sel(time=t0)        # whole grid, then slice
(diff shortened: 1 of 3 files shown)


benchmarks/geospatial/09_lazy_roundtrip.py
```diff
@@ -0,0 +1,143 @@
...
+#   "numpy",
+#   "pandas",
+#   "dask",
+#   "pooch",
+#   "netCDF4",
+# ]
+#
+# [tool.uv.sources]
+# xarray-sql = { path = "../../", editable = true }
+# ///
```
alxmrs, line 15: I see this as a good unit test or property that cross cuts all the other geo benchmarks, but I don't think it alone makes for a good benchmark example.

ghostiee-11: Okay, makes sense

benchmarks/geospatial/_harness.py
```diff
@@ -185,10 +185,12 @@ def measured(label: str) -> Iterator[None]:
...
-        yield
-        elapsed = time.perf_counter() - t0
-        _, peak = tracemalloc.get_traced_memory()
-        tracemalloc.stop()
+        try:
+            yield
+        finally:
+            elapsed = time.perf_counter() - t0
+            _, peak = tracemalloc.get_traced_memory()
+            tracemalloc.stop()
```
alxmrs, line 193: This fix seems like a good idea

benchmarks/geospatial/_harness.py
```diff
@@ -219,6 +221,16 @@ def assert_grid_close(
...
+    short = {
+        d: (got.sizes[d], ref.sizes[d])
+        for d in ref.dims
+        if d in got.sizes and got.sizes[d] != ref.sizes[d]
+    }
+    if short:
+        raise AssertionError(
+            f"{name}: SQL result does not cover the reference grid "
+            f"(dim: got vs ref = {short}); the comparison would be partial"
+        )
```
alxmrs, line 233: I'm surprised that Xarray's all close doesn't cover this case. Are you sure this is necessary?

alxmrs (commented):
A few other notes. If you removed case 09 and we got to the bottom of the fixes, I'd be happy to merge this.

benchmarks/geospatial/_harness.py
ghostiee-11: good catch, you're right allclose would catch it on its own. it's the `reindex_like(got)` one line up that hides it: it shrinks ref down to got's coords first, so a result missing cells still passes on the subset.

```python
got = ref.isel(lat=[0, 1, 2])  # 2 cells dropped
xr.testing.assert_allclose(got, ref.reindex_like(got))  # passes, silently
xr.testing.assert_allclose(got, ref)                    # raises
```

so the guard just restores the check `reindex_like` removes. could also drop `reindex_like` for `xr.align(..., join="exact")`, but that line handles label ordering so the guard felt smaller. either works.

alxmrs:
Closing as stale -- the changes that we want to keep I incorporated as review feedback. Thanks for this effort.

ghostiee-11:
Okay thanks Alex

PR #191, opened 2026-06-27
Add TEM benchmark (case 10): zonal means + eddy fluxes as a GROUP BY

ghostiee-11:
Adds `10_tem.py`: the Transformed Eulerian Mean (zonal means plus the eddy momentum, heat, and vertical fluxes) as one grouped aggregate, using the covariance identity `AVG(u*v) - AVG(u)*AVG(v)` so there is no self-join.
Reads u/v/T/w lazily from ARCO-ERA5 on three pressure levels and asserts the result matches the same diagnostic in pure xarray; the TEM workflow dcherian raised in #1545.

```text
  ⏱  SQL TEM (zonal means + eddy covariances, lazy read): 197.715s  (peak 892.3 MB)
  ⏱  xarray reference: 240.136s  (peak 395.0 MB)
  ✅ zonal-mean u (u_bar): SQL matches xarray reference (n=2,163, coordinate-aligned)
  ✅ eddy momentum flux u'v': SQL matches xarray reference (n=2,163, coordinate-aligned)
  ✅ eddy heat flux v't': SQL matches xarray reference (n=2,163, coordinate-aligned)
  ✅ vertical flux u'w': SQL matches xarray reference (n=2,163, coordinate-aligned)

  Result (SQL -> xarray):
  <xarray.Dataset> Size: 124kB
  Dimensions:   (time: 1, level: 3, latitude: 721)
  Coordinates:
    * time      (time) datetime64[ns] 2020-06-01T12:00:00
    * level     (level) int64 250 500 850
    * latitude  (latitude) float32 -90.0 -89.75 ... 89.75 90.0
  Data variables: u_bar, v_bar, t_bar, w_bar, upvp (u'v'), vptp (v't'), upwp (u'w')

  🎉 TEM: zonal means + eddy fluxes as GROUP BY (ARCO-ERA5): done.
```

### DIFF
--- a/benchmarks/geospatial/10_tem.py
+++ b/benchmarks/geospatial/10_tem.py
@@ -1,0 +1,164 @@
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#   "xarray-sql",
+#   "xarray",
+#   "gcsfs",
+#   "zarr>=3",
+# ]
+#
+# [tool.uv.sources]
+# xarray-sql = { path = "../../", editable = true }
+# ///
+"""Transformed Eulerian Mean: zonal means and eddy fluxes are a GROUP BY.
+
+The Transformed Eulerian Mean (TEM) is a standard atmospheric-circulation
+diagnostic: average the flow around each latitude circle, then measure how the
+departures from that average (the eddies) carry momentum and heat. In the array
+paradigm it is ``ds.mean("longitude")`` plus a few ``(x - x_bar)`` products
+averaged again over longitude.
+
+Every piece of that is relational. A zonal mean is ``GROUP BY latitude``
+collapsing longitude. An eddy flux such as the momentum flux
+``u'v' = mean((u - u_bar)(v - v_bar))`` is, by the covariance identity, just
+``AVG(u*v) - AVG(u)*AVG(v)``: one grouped pass, no self-join. So the whole
+diagnostic is::
+
+    SELECT time, level, latitude,
+           AVG(u) AS u_bar, ...,
+           AVG(u*v) - AVG(u)*AVG(v) AS upvp,   -- eddy momentum flux u'v'
+           AVG(v*t) - AVG(v)*AVG(t) AS vptp,   -- eddy heat flux v't'
+           AVG(u*w) - AVG(u)*AVG(w) AS upwp    -- vertical momentum flux u'w'
+    FROM era5 GROUP BY time, level, latitude
+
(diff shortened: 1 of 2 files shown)


benchmarks/geospatial/10_tem.py
```diff
@@ -0,0 +1,164 @@
...
+
+    SELECT time, level, latitude,
+           AVG(u) AS u_bar, ...,
+           AVG(u*v) - AVG(u)*AVG(v) AS upvp,   -- eddy momentum flux u'v'
+           AVG(v*t) - AVG(v)*AVG(t) AS vptp,   -- eddy heat flux v't'
+           AVG(u*w) - AVG(u)*AVG(w) AS upwp    -- vertical momentum flux u'w'
+    FROM era5 GROUP BY time, level, latitude
+
+This is the diagnostic dcherian raised in the Large Scale Geospatial Benchmarks
+discussion (coiled/benchmarks #1545); the SQL reads like its textbook definition.
```
alxmrs, line 35: That this reads like a textbook definition is right and also makes me curious: could we do calculus in SQL? Check out this example: https://xgcm.readthedocs.io/en/latest/xgcm-examples/01_eccov4/#spatial-derivatives-heat-budget

(Background: General Circulation Models are how ERA5 and other similar climate simulations are made.)

This is a crazy reach project, but in the future: could we do the core operations that justified xGCM's creation in SQL via this data model? If we could create realistic examples (especially, ones that could scale with distributed SQL), then I think it would warrant at least a blog post if not a Pangeo Showcase presentation. 

https://mitgcm.readthedocs.io/en/latest/examples/examples.html

benchmarks/geospatial/10_tem.py
```diff
@@ -0,0 +1,164 @@
...
+            "era5",
+            ds,
+            chunks={"time": 1},
+            table_names={
+                ("time", "latitude", "longitude"): "surface",
+                ("time", "level", "latitude", "longitude"): "atmosphere",
+            },
+        )
+
+    sql = """
```
alxmrs, line 99: This is beautiful

alxmrs: I didn't realize it would be so readable.

alxmrs: Honestly, I find the Xarray version to be more confusing.

alxmrs (approved):
Beautiful. Thanks, this is a great example. I'll run perf tests of this later.

PR #193, opened 2026-06-28
feat: skip per-dim discovery scans when query is an unfiltered scan

### LINKED ISSUE #171: to_dataset should preserve source dim order without test-side sortby
Right now `XarrayDataFrame.to_dataset()` discovers per-dim coords via `inner_df.select(col(d)).distinct().sort(col(d).sort())`, which always returns ascending values. Source Datasets whose coords are descending (e.g. `air_temperature.lat` runs 75 -> 15) round-trip with the lat axis flipped, and the round-trip tests have to call `sortby` on both sides to compare. Same for arbitrary user ordering of integer coords.

The fix is to derive coord order from the registered template when one is available, and only fall back to `SELECT DISTINCT ... ORDER BY` when there is no template. We would also stop sorting on dims the template already covers.

Tracking this so we can drop the `sortby` workarounds in `tests/test_ds.py` (`test_round_trip_identity`, `test_lazy_isel_*`, `test_lazy_compute_returns_eager`) and so users get back exactly what they registered.

Spun out of #167.

ghostiee-11:
For an unfiltered scan over a single registered table (Projection / Sort / TableScan / SubqueryAlias only) the result's coord extent is exactly the source's. `_build_lazy_scan` now reuses the registered Dataset's coord arrays directly instead of running one DISTINCT scan per dim. Filtered / aggregated / joined / limited / multi-table queries fall back to the existing discovery path unchanged. Also closes #171 (lazy round-trip now preserves source dim order, e.g. descending lat).

## Results

`to_dataset()` construction on NCEP `air_temperature` (chunks={'time':24}):

| stage | before | after |
|---|---:|---:|
| peak memory | 77 MB | **0.1 MB** |
| wall time   | 3.4 s | **7 ms**  |

171 / 171 tests pass; detector audited across 8 query shapes.

### DIFF
--- a/tests/test_ds.py
+++ b/tests/test_ds.py
@@ -202,6 +202,110 @@ def test_order_by_direction_sets_dim_order(air_dataset_small):
     np.testing.assert_allclose(out["air_avg"].values, expected.values)
 
 
+def test_unfiltered_scan_skips_dim_discovery(air_dataset_small):
+    """``SELECT * FROM <registered>`` does no per-dim discovery scans.
+
+    The pre-fix :func:`_build_lazy_scan` ran one full source scan per dim
+    via ``inner_df.select(col(d)).distinct().sort(...)`` -- a single-column
+    projection per dim, multiplied across the dim count. With a registered
+    template and an unfiltered plan, :func:`_maybe_template_coords` returns
+    the template's coord arrays directly and the discovery loop is skipped
+    entirely. Verified by the iteration callback's projection list: no read
+    where the projection is a single dim column.
+    """
+    from xarray_sql.reader import read_xarray_table
+
+    reads: list = []
+    table = read_xarray_table(
+        air_dataset_small,
+        chunks={"time": 6},
+        _iteration_callback=lambda block, proj: reads.append(list(proj)),
+    )
+    ctx = XarrayContext()
+    ctx.register_table("air", table)
+    ctx._registered_datasets["air"] = air_dataset_small
+
+    reads.clear()
+    out = ctx.sql('SELECT * FROM "air"').to_dataset()
+    single_dim_reads = [
+        proj for proj in reads if set(proj) <= {"time", "lat", "lon"}
+    ]
+    assert not single_dim_reads, (
+        f"fast path should skip per-dim discovery scans, got "
+        f"{single_dim_reads!r}"
+    )
+    # The fast path must not break correctness on first access.
+    out.compute()  # safety: the Dataset still resolves
+
+
+def test_filtered_scan_still_uses_dim_discovery(air_dataset_small):
+    """A ``WHERE`` clause forbids the template-coord fast path.
+
+    With a filter the result's coord extent is a subset of the source, so
+    reusing template coords would silently widen the output. The dispatch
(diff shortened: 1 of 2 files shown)


tests/test_ds.py
```diff
@@ -202,6 +202,110 @@ def test_order_by_direction_sets_dim_order(air_dataset_small):
...
+    table = read_xarray_table(
+        air_dataset_small,
+        chunks={"time": 6},
+        _iteration_callback=lambda block, proj: reads.append(list(proj)),
+    )
+    ctx = XarrayContext()
+    ctx.register_table("air", table)
+    ctx._registered_datasets["air"] = air_dataset_small
+
+    reads.clear()
```
alxmrs, line 228: What does the clear do?

tests/test_ds.py
```diff
@@ -202,6 +202,110 @@ def test_order_by_direction_sets_dim_order(air_dataset_small):
...
+    in :func:`_maybe_template_coords` falls back to per-dim discovery in
+    that case, which is observable as a positive read count on the source.
+    """
+    from xarray_sql.reader import read_xarray_table
+
+    reads: list = []
+    table = read_xarray_table(
+        air_dataset_small,
+        chunks={"time": 6},
+        _iteration_callback=lambda block, proj: reads.append(block),
```
alxmrs, line 255: Can we make this append consistent with the above version (block or proj)

tests/test_ds.py
```diff
@@ -202,6 +202,110 @@ def test_order_by_direction_sets_dim_order(air_dataset_small):
...
+    output if their coord ranges differ.
+    """
+    other = air_dataset_small.isel(lat=slice(0, 5))
+    assert other.sizes["lat"] != air_dataset_small.sizes["lat"]
+    ctx = XarrayContext()
+    ctx.from_dataset("air", air_dataset_small, chunks={"time": 24})
+    ctx.from_dataset("other", other, chunks={"time": 24})
+    out = ctx.sql('SELECT * FROM "air"').to_dataset(
+        dims=["time", "lat", "lon"], template=other
+    )
```
alxmrs, line 284: Cool, good test

tests/test_ds.py
```diff
@@ -202,6 +202,110 @@ def test_order_by_direction_sets_dim_order(air_dataset_small):
...
+    # Scanned table is "air", so lat must match air's full lat axis.
+    np.testing.assert_array_equal(
+        out["lat"].values, air_dataset_small["lat"].values
+    )
+
+
+def test_round_trip_preserves_descending_lat_on_lazy_path(air_dataset_small):
```
alxmrs, line 291: Nice

xarray_sql/ds.py
```diff
@@ -432,33 +432,115 @@ def _materialize(
     return xr.Dataset(data_vars=data_vars, coords=coords_arg)
 
 
+_PURE_SCAN_NODES = {"Projection", "Sort", "TableScan", "SubqueryAlias"}
+
+
+def _unfiltered_scan_table(inner_df: Any) -> str | None:
```
alxmrs, line 438: Nice

alxmrs (approved):
A few notes -- great idea. I can't wait to benchmark this optimization.

Issue #194, opened 2026-06-28
Support a mutable/appendable table for in-SQL iterative state (e.g. optimizer trajectories)


alxmrs:
## Motivation

The autograd work (#192) makes it possible to run **gradient descent in SQL** — the gradient of a loss is `AVG(grad(loss, θ))`, and an update is `θ_new = θ - lr * AVG(grad(loss, θ))`. A natural, SQL-native way to express the optimizer is a **parameter table that grows one row per step**, so the whole training trajectory becomes a relation you can query (e.g. the loss curve in a single `GROUP BY`). See the discussion in https://github.com/alxmrs/xarray-sql/pull/192#discussion_r3488049015.

Today that "append a new generation" step can't happen in SQL. The tables registered via `XarrayContext.from_dataset` are backed by a **read-only, lazy** Arrow stream provider (`LazyArrowStreamTable`), so there is no `INSERT INTO`. The demo works around it by rebuilding the whole history as an xarray `Dataset` and **re-registering** it each step:

```python
def register_params():
    ctx.deregister_table("params")
    ds = xr.Dataset({"a": (("step",), a_hist), "b": (("step",), b_hist)},
                    coords={"step": steps})
    ctx.from_dataset("params", ds, chunks={"step": len(steps)})
```

It's correct and fine for a demo, but it's the one un-SQL-native part of an otherwise fully-relational training loop, and it re-materializes O(history) each step.

## What we'd want

Some way to do, in SQL:

```sql
INSERT INTO params
SELECT :next_step,
       cur.a - lr * AVG(grad(loss, a)),
       cur.b - lr * AVG(grad(loss, b))
FROM d CROSS JOIN (SELECT a, b FROM params WHERE step = :k) cur
GROUP BY cur.a, cur.b;
```

so the optimizer's state lives in the engine and the Python side only drives iteration (or, eventually, doesn't — see "recursive CTE" below).

## Design tension (why "xarray-backed append" specifically is awkward)

The obvious framing — "make xarray tables appendable" — fights the data model:

- xarray/Zarr tables model **immutable, dense, gridded arrays**. A SQL row is one cell `(coords…, values…)`. "Appending a row" to a gridded array isn't an `INSERT`; it's growing/concatenating along a coordinate dimension (a rechunk/`xr.concat`), which only makes sense for some tables (a 1-D `params(step)`) and not others (a 3-D `(time, lat, lon)` field).
[...]

PR #195, opened 2026-06-28
Demos: differentiable SQL over ARCO-ERA5 and gradient descent in SQL

alxmrs:
Stacked on the autograd feature branch (#192) — runnable benchmark scripts kept out of the core branch so it stays reviewable.

## What this adds (`benchmarks/`)

- **`grad_era5.py`** — the autograd feature on real climate data. A physical quantity is written as an analytic SQL formula over ARCO-ERA5 variables, and `grad(...)` differentiates it **symbolically**, per grid cell (the relational `vmap(grad(f))`). Two cases, each checked against an analytic reference:
  - wind speed `sqrt(u² + v²)` → `grad(speed, u) = u/speed` (exact)
  - saturation vapour pressure `A·exp(B·tc/(tc+C))` → `grad(e_s, T)` vs the closed-form Clausius–Clapeyron slope

  Reads ARCO-ERA5 anonymously from GCS (needs `gcsfs` + network); each query round-trips back to an `xarray.Dataset`.

- **`grad_descent.py`** — gradient descent in SQL. Differentiating through an aggregate is linearity, so the gradient is `AVG(grad(loss, θ))`. The optimiser trajectory is a `params(step, a, b)` table that grows one generation per step, with the update computed in SQL (`new_a = a - lr*AVG(grad(loss, a))`); the whole loss curve is one `GROUP BY` over the history. Fit matches numpy least-squares. Self-contained (no network).

- **`benchmarks/README.md`** describing both.

Each script declares its deps inline (PEP 723) and runs with `uv run benchmarks/<name>.py`; nothing is added to `pyproject.toml`, and the suite lives outside the package so it isn't collected by pytest or shipped in the wheel.

---
_Generated by [Claude Code](https://claude.ai/code/session_017mDoFJgsm9kS7SicGoCVF6)_

### DIFF
--- a/benchmarks/grad_descent.py
+++ b/benchmarks/grad_descent.py
@@ -1,0 +1,115 @@
+# /// script
+# requires-python = ">=3.10"
+# dependencies = [
+#   "xarray_sql",
+#   "xarray",
+#   "numpy",
+# ]
+#
+# [tool.uv.sources]
+# xarray_sql = { path = "..", editable = true }
+# ///
+"""Gradient descent as a single declarative SQL query.
+
+Fits a line ``y ~= a*x + b`` by minimising the mean squared error — with the
+**entire training loop expressed as one recursive CTE**, no Python iteration.
+
+Two pieces:
+
+1. **grad compiles the update rule.** ``differentiate_sql`` turns the per-row
+   loss into the symbolic derivative *as SQL text* — the autograd engine acting
+   as a calculus compiler:
+
+       da = differentiate_sql("(y-(a*x+b))^2", "a")   # -> "-2*((a*x+b)-y)*x", etc.
+
+2. **A recursive CTE is the optimiser.** ``params(step, a, b)`` starts at one
+   row and each recursion appends the next generation, descending along the
+   gradient (``AVG`` of the compiled rule over the data):
+
+       params.a - lr * AVG(da)
+
+   So the whole loop — gradient, update, and iteration — is declarative SQL;
+   the optimisation trajectory is the rows of one query.
+
+Why two pieces instead of ``grad(...)`` directly inside the recursion? ``grad``
+needs the Substrait round-trip, and Substrait has no recursion — so ``grad``
+can't live inside a recursive CTE (tracked in #194 / a follow-up). Differentiating
+once to plain SQL sidesteps that: the recursive query contains no ``grad`` marker.
+
+Run standalone:
+
+    uv run benchmarks/grad_descent.py
+"""
+
+from __future__ import annotations
+
+import numpy as np
+import xarray as xr
+
+import xarray_sql as xql
+
+# Per-row loss r^2 with residual r = y - (a*x + b), over columns a, b, x, y.
+RESIDUAL = "(y - (a * x + b))"
+LOSS = f"{RESIDUAL} * {RESIDUAL}"
+COLUMNS = ["a", "b", "x", "y"]
+LR = 0.4
+STEPS = 200
+
+
+def main() -> None:
+    rng = np.random.default_rng(0)
+    n = 500
+    x = rng.uniform(0.0, 1.0, n)
+    a_true, b_true = 2.0, -1.0
+    y = a_true * x + b_true + rng.normal(0.0, 0.01, n)
+
+    ctx = xql.XarrayContext()
+    ctx.from_dataset(
+        "d",
+        xr.Dataset(
+            {"x": (("i",), x), "y": (("i",), y)}, coords={"i": np.arange(n)}
+        ),
+        chunks={"i": n},
+    )
+
+    # grad compiles the per-row update rule to SQL, once.
+    da = xql.differentiate_sql(LOSS, "a", COLUMNS)
+    db = xql.differentiate_sql(LOSS, "b", COLUMNS)
+    print(f"d(loss)/da = {da}")
+    print(f"d(loss)/db = {db}\n")
+
+    # The entire training loop is one declarative recursive query: each step
+    # appends the next generation, descending along the SQL-computed gradient.
+    trajectory = ctx.sql(
+        f"""
+        WITH RECURSIVE params(step, a, b) AS (
+          SELECT 0 AS step, CAST(0.0 AS DOUBLE) AS a, CAST(0.0 AS DOUBLE) AS b
+          UNION ALL
+          SELECT params.step + 1               AS step,
+                 params.a - {LR} * AVG({da})    AS a,
+                 params.b - {LR} * AVG({db})    AS b
+          FROM params CROSS JOIN d
+          WHERE params.step < {STEPS}
+          GROUP BY params.step, params.a, params.b
+        )
+        SELECT step, a, b FROM params ORDER BY step
+        """
+    ).to_pandas()
+
+    print("trajectory (every 40th generation):")
+    print(trajectory.iloc[::40].to_string(index=False))
+
+    a, b = float(trajectory["a"].iloc[-1]), float(trajectory["b"].iloc[-1])
+    a_ols, b_ols = np.polyfit(x, y, 1)
+    print(
+        f"\nSQL gradient descent:  a={a:.4f}  b={b:.4f}  ({len(trajectory)} generations)"
+    )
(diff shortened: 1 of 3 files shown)


benchmarks/grad_descent.py
```diff
@@ -0,0 +1,124 @@
...
+        GROUP BY cur.a, cur.b
+        """
+
+    for k in range(STEPS):
+        register_params()
+        row = ctx.sql(update_sql(k)).to_pandas()
+        steps.append(k + 1)
+        a_hist.append(float(row["a"][0]))
+        b_hist.append(float(row["b"][0]))
+    register_params()
```
alxmrs: I believe we can do this iteration in SQL too -- these just add more rows into a (train?) table.

alxmrs: Love this — done. `params` is now a DataFusion in-memory table and each step is literally "add a row":

```sql
INSERT INTO params VALUES (:next_step, :new_a, :new_b)
```

where `(new_a, new_b)` come from the SQL update `a - lr*AVG(grad(loss, a))`. So the gradient, the update, *and* the append all happen in SQL — a training step is an `INSERT` into the table, and the loss curve is one `GROUP BY` over the accumulated history. (This needed a small core change: the rewrite now resolves `grad` over any registered table, not just xarray-backed ones — see #192.)

One honest seam remains: the Python `for` loop still *sequences* the steps. Doing the iteration itself in SQL — no Python loop — would be a recursive CTE, which DataFusion/Substrait don't round-trip yet (tracked in #194). So today it's "each step is SQL `INSERT`", not "the whole loop is one SQL statement". The `INSERT ... SELECT grad(...)` single-statement form also doesn't go through the Substrait rewrite (DML plans don't convert), which is why the update is a `SELECT` feeding an `INSERT VALUES`.

---

benchmarks/grad_descent.py
```diff
@@ -0,0 +1,120 @@
...
+        # the SQL-computed gradient...
+        row = ctx.sql(
+            f"""
+            WITH cur AS (SELECT a, b FROM params WHERE step = {k})
+            SELECT cur.a - {LR} * AVG(grad({LOSS}, a)) AS a,
+                   cur.b - {LR} * AVG(grad({LOSS}, b)) AS b
+            FROM d CROSS JOIN cur
+            GROUP BY cur.a, cur.b
+            """
+        ).to_pandas()
```
alxmrs: Pandas invokes a copy, let's keep it all in datafusion (lazily evaluated).

benchmarks/grad_descent.py
```diff
@@ -0,0 +1,120 @@
...
+        # ...and append it to the parameter table with a SQL INSERT.
+        na, nb = float(row["a"][0]), float(row["b"][0])
+        ctx.sql(
+            f"INSERT INTO params VALUES ({k + 1}, {na!r}, {nb!r})"
+        ).collect()
```
alxmrs: Can't we select over a range of STEPSs values (iteration) and do this in one query as a set of nested SELECT statements? Why do we need to spill into Python and use the imperative paradigm? I think the entire training loop could be declarative tables.

benchmarks/grad_era5.py
```diff
@@ -0,0 +1,180 @@
...
+          latitude,
+          longitude,
+          {speed}            AS wind_speed,
+          grad({speed}, {U}) AS d_speed_d_u,
+          grad({speed}, {V}) AS d_speed_d_v
+        FROM era5
+        """
+        )
+        .to_dataset(dims=["latitude", "longitude"])
+        .sortby(_SORT)
```
alxmrs: This can be done with an ORDER BY SQL expr.

benchmarks/grad_era5.py
```diff
@@ -0,0 +1,180 @@
...
+            SELECT
+              latitude,
+              longitude,
+              {es}            AS e_s,
+              grad({es}, {T}) AS de_s_dt
+            FROM era5
+            """
+        )
+        .to_dataset(dims=["latitude", "longitude"])
+        .sortby(_SORT)
```
alxmrs: Same here.

alxmrs (commented):
More feedback, this time on the ERA5 example.

benchmarks/grad_descent.py
alxmrs: Done — the whole training loop is now **one declarative recursive CTE**, no Python iteration:

```sql
WITH RECURSIVE params(step, a, b) AS (
  SELECT 0, 0.0, 0.0
  UNION ALL
  SELECT params.step + 1, params.a - lr*AVG(da), params.b - lr*AVG(db)
  FROM params CROSS JOIN d WHERE params.step < STEPS
  GROUP BY params.step, params.a, params.b)
SELECT * FROM params ORDER BY step
```

The one wrinkle: `grad()` can't appear *inside* the recursion, because it needs the Substrait round-trip and Substrait has no recursion (`to_substrait_plan` rejects a `RecursiveQuery`). So I split it: a new engine helper `xql.differentiate_sql(loss, "a", cols)` differentiates the per-row loss to SQL text **once** (autograd as a "compiler"), and the recursive CTE iterates that plain-SQL rule. `da`/`db` are the compiled gradients; gradient + update + iteration are all in SQL. Fits to the OLS solution exactly. (`differentiate_sql` is in #192.)

---

benchmarks/grad_descent.py
alxmrs: Resolved by the recursive-CTE rewrite (see the thread below): the training loop is now a single query, so there's no per-step `to_pandas()` — just one `.to_pandas()` at the very end to print the trajectory/report the fit.

---

benchmarks/grad_era5.py
alxmrs: Done — the query now ends with `ORDER BY latitude, longitude` and I dropped the Python `.sortby` on the SQL output. (The reference array is still sorted in Python, since it comes from xarray in ERA5's native descending-latitude order.)

---

benchmarks/grad_era5.py
alxmrs: Done here too — `ORDER BY latitude, longitude` in the Clausius–Clapeyron query, output `.sortby` removed.

---

benchmarks/grad_era5.py
```diff
@@ -0,0 +1,175 @@
...
+
+    # Reference in float64 (the columns are float32): the exact derivative is
+    #   d(e_s)/dT = e_s * B*C / (tc + C)^2.
+    temp = ref["2m_temperature"].astype("float64")
+    tc_ref = temp - 273.15
+    es_ref = a * np.exp(b * tc_ref / (tc_ref + c))
+    des_dt_ref = es_ref * (b * c) / (tc_ref + c) ** 2
+
+    xr.testing.assert_allclose(
+        out["e_s"], es_ref.sortby(_SORT).rename("e_s"), rtol=1e-5
```
alxmrs: You can eliminate the ref sortby id you order latitude by DESC in the SQL query.

benchmarks/grad_era5.py
```diff
@@ -0,0 +1,175 @@
...
+    )
+    xr.testing.assert_allclose(
+        out["de_s_dt"], des_dt_ref.sortby(_SORT).rename("de_s_dt"), rtol=1e-5
```
alxmrs: Same here.

benchmarks/grad_era5.py
alxmrs: Done — both queries now `ORDER BY latitude DESC, longitude` to match ERA5's native order, and **both reference `.sortby` calls are gone** (the `_SORT` helper too). No sorting on either side now.

One subtlety I hit making this work: `to_dataset` only preserves the SQL `ORDER BY` order for a **single partition**. Across multiple partitions it reconstructs coordinates ascending regardless (a `latitude DESC` query still came back ascending). The demo block is tiny, so I switched it to one partition (`block.chunk()`) and noted why. The general fix — preserve source dim order across partitions on round-trip — is exactly #193; once that lands this works for chunked tables too.

---

benchmarks/grad_era5.py
alxmrs: Done — same change in the Clausius–Clapeyron query (`ORDER BY latitude DESC, longitude`, reference sort removed).

---

PR #199, opened 2026-06-29
test: peak-memory integration tests for the lazy round-trip

### LINKED ISSUE #172: Measure to_dataset peak memory + wall time on non-toy data
The lazy-backend memory claim (only the requested region is materialized) is verified today only against the conftest fixtures and one xarray tutorial dataset. Those are all small enough that a regression that quietly buffers the whole result would still fit in RAM and look fine.

What we want is a benchmark / smoke test that runs on a real-world-sized Dataset (a few GB on disk, multiple chunks) and asserts:

- peak RSS for a single-region access stays bounded by chunk size, not result size,
- streaming aggregation does not blow up while reducing a long axis,
- `compute()` time scales linearly in materialized rows.

Likely lives under `perf_tests/`, gated behind a marker so CI does not always pay the download cost.

Spun out of #167.

ghostiee-11:
Closes #172.

Two opt-in tests behind the existing `integration` marker, registered in `pyproject.toml` (and default suite now uses `-m "not integration"` so plain `pytest` skips them, matching the CI workflow's existing `pytest -v . -m "not integration"`):

- A single `.sel(time=t0).load()` peaks well below the eager whole-grid materialization (lazy < 10 MB, eager > 50 MB on `air_temperature`, ratio > 10x).
- A `GROUP BY` reduction along the long axis stays under 4x the source's nominal size, with values matching `mean(dim='time')`.

Run on demand: `pytest -m integration`.

### DIFF
--- a/tests/test_to_dataset_perf.py
+++ b/tests/test_to_dataset_perf.py
@@ -1,9 +1,8 @@
-"""Opt-in peak-memory tests for the lazy SQL -> xarray round-trip.
+"""Peak-memory tests for the lazy SQL -> xarray round-trip.
 
-Asserts the lazy backend honors its contract: a single-slab access
+Asserts the lazy backend honors its contract: a single-chunk access
 peaks far below an eager whole-grid materialization, and a streaming
-aggregation does not balloon past the source size. Gated by the
-``integration`` marker; run with ``pytest -m integration``.
+aggregation does not balloon past the source size.
 """
 
 import gc
@@ -14,8 +13,6 @@ import pytest
 import xarray as xr
 
 from xarray_sql import XarrayContext
-
-pytestmark = pytest.mark.integration
 
 
 def _peak_mb(fn):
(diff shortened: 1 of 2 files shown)


tests/test_to_dataset_perf.py
```diff
@@ -0,0 +1,118 @@
...
+    Module-scoped so the pooch download (~3.5 MB) is amortized across
+    tests in this file. Skips when the tutorial dataset is unreachable.
+    """
+    try:
+        return xr.tutorial.open_dataset("air_temperature").chunk({"time": 24})
+    except (OSError, ValueError, ImportError) as e:
+        pytest.skip(f"air_temperature tutorial dataset unavailable: {e}")
+
+
+def test_lazy_slab_peak_memory_is_bounded(air_source):
```
alxmrs: Let's not use the term "slab". Claude does like it though.

tests/test_to_dataset_perf.py
```diff
@@ -0,0 +1,118 @@
...
+    """``.sel(time=t0).load()`` materializes only the slab, not the cube.
```
alxmrs: This is a single chunk.

tests/test_to_dataset_perf.py
```diff
@@ -0,0 +1,118 @@
...
+import gc
+import tracemalloc
+
+import numpy as np
+import pytest
+import xarray as xr
+
+from xarray_sql import XarrayContext
+
+pytestmark = pytest.mark.integration
```
alxmrs: Does this mean the whole file is market as an integration test?

alxmrs (changes requested):
Naming issue, then this test should be good to go. 

Frankly, if we apply this to tutorial datasets (which are small), then these could be unit tests.

tests/test_to_dataset_perf.py
ghostiee-11: Yeah claude does this

alxmrs (approved):

PR #201, opened 2026-06-30
Exact table statistics for the optimizer (DataFusion 54)

alxmrs:
Report exact statistics from the xarray scan so DataFusion's cost-based optimizer can plan joins and aggregations well — on the ordinary engine, with no second engine and the full `datafusion-python` DataFrame surface preserved.

## Why this works now

Earlier exploration went down a "native in-process engine" path because `datafusion-ffi` (52/53) dropped `Statistics` across the FFI boundary — a foreign scan reported unknown cardinality. **DataFusion 54 forwards `ExecutionPlan` statistics across FFI** (`FFI_ExecutionPlan` gained `partition_statistics`), so the statistics our scan reports now reach the optimizer through the normal FFI table-provider path. The native engine is gone.

## What's here

- **`XarrayScanExec`** wraps the `StreamingTableExec` from `scan()` and reports exact `Statistics`. Every stat is derived from coordinate metadata xarray already knows — none of it scans the data — and each is exact, not an estimate:
  - `num_rows` — summed product of each surviving chunk's dimension sizes. Drives `JoinSelection`'s build-side choice and lets `COUNT(*)` skip the scan.
  - `total_byte_size` — `num_rows × fixed row width` (derived in Rust from the projected schema's primitive widths).
  - per dimension-coordinate column — exact `min`/`max` (the join/filter keys) and `null_count = 0` (grid axes are always fully populated). Data variables are left unknown; their bounds would require a scan.
- **Why the wrapper (and not just `TableProvider::statistics()`)**: the physical `JoinSelection` rule reads statistics off the `ExecutionPlan`, not off the `TableProvider` — so injecting them at the plan node is what actually drives build-side selection, even under DF 54's FFI forwarding. Confirmed empirically: `TableProvider::statistics()` alone leaves the join as `mode=Partitioned`.
- **Where the numbers come from**: `num_rows` is plumbed from Python as an optional third tuple element `(factory, metadata, num_rows)` — `_block_len(block)`, the product of the chunk's slice extents (the 2-tuple form still works). The min/max bounds reuse the same coordinate metadata already computed for partition pruning. Nothing is recomputed from the data.
- **Dependency upgrade**: `datafusion` + `datafusion-ffi` 52 → 54 (and `arrow` 57 → 58, `pyo3` 0.26 → 0.28 to match), plus the `datafusion` Python dep → 54.

## Verified

- A big-vs-small join plans as `HashJoinExec: mode=CollectLeft` with the small side's `Rows=Exact(64)` carried through the FFI boundary (`FFI_ExecutionPlan: XarrayScanExec`) — the cost-based build-side choice statistics unlock.
- A projected scan's `EXPLAIN` shows `Rows=Exact(2000), Bytes=Exact(48000), [(Col[0]: Min=Exact(Int64(0)) Max=Exact(Int64(3)) Null=Exact(0)), …]`, with data-variable columns carrying no bounds.
- `COUNT(*)` is answered from the exact statistics without scanning.
- Full suite green (`tests/test_stats.py` pins the statistics contract; reader tests that used `COUNT(*)` to force a scan now use `SELECT *`, since `COUNT(*)` is metadata-only once statistics are exact).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,20 +41,22 @@
 //! Will skip loading partitions whose time ranges are entirely before 2020-02-01.
 //! Supported operators: `=`, `<`, `>`, `<=`, `>=`, `BETWEEN`, `IN`, `AND`, `OR`.
 
-use std::any::Any;
 use std::collections::{HashMap, HashSet};
 use std::ffi::CString;
 use std::fmt::Debug;
 use std::sync::Arc;
 
 use arrow::array::RecordBatch;
-use arrow::datatypes::{Schema, SchemaRef};
-use arrow::pyarrow::FromPyArrow;
+use arrow::datatypes::{DataType, Schema, SchemaRef, TimeUnit};
+use arrow::pyarrow::{FromPyArrow, ToPyArrow};
 use async_stream::try_stream;
 use async_trait::async_trait;
 use datafusion::catalog::streaming::StreamingTable;
 use datafusion::catalog::Session;
-use datafusion::common::{DataFusionError, Result as DFResult, ScalarValue};
+use datafusion::common::stats::Precision;
+use datafusion::common::{
+    ColumnStatistics, DataFusionError, Result as DFResult, ScalarValue, Statistics,
+};
 use datafusion::datasource::TableProvider;
 use datafusion::execution::TaskContext;
 use datafusion::logical_expr::expr::InList;
@@ -63,7 +65,9 @@ use datafusion::logical_expr::{
 };
 use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
 use datafusion::physical_plan::streaming::PartitionStream;
-use datafusion::physical_plan::{ExecutionPlan, SendableRecordBatchStream};
+use datafusion::physical_plan::{
+    DisplayAs, DisplayFormatType, ExecutionPlan, PlanProperties, SendableRecordBatchStream,
+};
 use datafusion_ffi::proto::logical_extension_codec::FFI_LogicalExtensionCodec;
 use datafusion_ffi::table_provider::FFI_TableProvider;
 use pyo3::prelude::*;
@@ -132,10 +136,18 @@ pub struct DimensionRange {
 }
 
 /// Metadata for a single partition, used for filter-based pruning.
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub struct PartitionMetadata {
     /// Dimension ranges for this partition, keyed by column name
     pub ranges: HashMap<String, DimensionRange>,
+    /// Exact number of rows in this partition (product of the chunk's
+    /// per-dimension sizes). Every partition supplies it, so the scan can
+    /// report exact `Statistics::num_rows` to the optimizer and cost-based
+    /// rules (join build-side selection, broadcast vs. shuffle) have real
+    /// cardinalities instead of guesses. xarray knows this exactly — it is
+    /// the product of the partition's dimension lengths — so unlike most
+    /// table providers these statistics are not estimates.
+    pub num_rows: usize,
 }
 
 impl PartitionMetadata {
@@ -485,12 +497,15 @@ fn python_to_scalar_bound(obj: &Bound<'_, PyAny>, dtype_tag: &str) -> PyResult<S
     }
 }
 
-/// Convert a bound Python metadata dict to Rust PartitionMetadata.
+/// Convert a bound Python metadata dict into per-dimension coordinate ranges.
 ///
 /// Operates on an already-bound reference so no additional GIL acquisition
 /// is needed — this is called from within a `#[pymethods]` context where
-/// the GIL is already held.
-fn convert_python_metadata_from_bound(meta_obj: &Bound<'_, PyAny>) -> PyResult<PartitionMetadata> {
+/// the GIL is already held. The caller pairs the result with the partition's
+/// row count to build a [`PartitionMetadata`].
+fn convert_python_ranges_from_bound(
+    meta_obj: &Bound<'_, PyAny>,
+) -> PyResult<HashMap<String, DimensionRange>> {
     type MetaDict = HashMap<String, (Py<PyAny>, Py<PyAny>, String)>;
     let meta_dict: MetaDict = meta_obj.extract()?;
     let py = meta_obj.py();
@@ -507,7 +522,7 @@ fn convert_python_metadata_from_bound(meta_obj: &Bound<'_, PyAny>) -> PyResult<P
             },
         );
     }
-    Ok(PartitionMetadata { ranges })
+    Ok(ranges)
 }
 
 impl Debug for PrunableStreamingTable {
@@ -522,10 +537,6 @@ impl Debug for PrunableStreamingTable {
 
 #[async_trait]
 impl TableProvider for PrunableStreamingTable {
-    fn as_any(&self) -> &dyn Any {
-        self
-    }
-
     fn schema(&self) -> SchemaRef {
         Arc::clone(&self.schema)
     }
@@ -563,10 +574,24 @@ impl TableProvider for PrunableStreamingTable {
         // Prune partitions based on filters
         let included_indices = self.prune_partitions(filters);
 
+        // Exact per-partition row counts for the partitions that survive
+        // pruning, in scan order. These feed `XarrayScanExec`'s statistics so
+        // the optimizer sees real cardinalities.
+        let included_metas: Vec<&PartitionMetadata> = included_indices
+            .iter()
+            .map(|&idx| &self.partitions[idx].1)
+            .collect();
+        let partition_rows: Vec<Precision<usize>> = included_metas
+            .iter()
+            .map(|meta| Precision::Exact(meta.num_rows))
+            .collect();
+
         // Handle empty case — all partitions pruned, return empty plan
         if included_indices.is_empty() {
             let empty_table = StreamingTable::try_new(Arc::clone(&self.schema), vec![])?;
-            return empty_table.scan(state, projection, filters, limit).await;
+            let inner = empty_table.scan(state, projection, filters, limit).await?;
+            let stats = build_scan_statistics(inner.schema().as_ref(), &included_metas);
+            return Ok(Arc::new(XarrayScanExec::new(inner, stats, partition_rows)));
         }
 
         // Determine whether to push projection down to the Python factory.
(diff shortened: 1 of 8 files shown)


xarray_sql/sql.py
```diff
@@ -10,9 +11,27 @@
...
+
+    * ``"native"``: tables are registered into an in-process DataFusion
+      ``SessionContext`` compiled into this extension, *bypassing the FFI
+      boundary*. The boundary drops table statistics and dynamic-filter
+      pushdown, so only the native engine can give the cost-based optimizer
+      exact cardinalities (for join build-side selection and broadcast-vs-
+      shuffle) and, in future, join-driven partition pruning. The native
+      engine is **eager** (it materialises the result) and does not register
+      Python scalar UDFs; use the default engine for ``cftime`` filters or
+      multi-dimension-group datasets.
```
alxmrs: That the native engine is eager and not lazy is a merge blocker. Laziness is a must.

xarray_sql/sql.py
```diff
@@ -102,6 +132,14 @@ def from_dataset(
                 name, input_table, chunks, coord_arrays=coord_arrays
             )
 
+        if self._native is not None:
+            raise NotImplementedError(
+                "The native engine does not yet support datasets whose "
+                "variables span multiple dimension groups (which register as "
+                f"a SQL namespace). {name!r} has groups {list(groups)}. Use "
+                "engine='ffi' for namespaced datasets."
+            )
```
alxmrs: We also needed this support.

xarray_sql/sql.py
```diff
@@ -174,9 +220,43 @@ def sql(self, query: str, *args, **kwargs) -> XarrayDataFrame:
...
+        if batches:
+            table = pa.Table.from_batches(batches)
+        else:
+            # Empty result: recover the schema so the wrapper still has columns.
+            schema = self._native.sql_schema(query)
+            table = pa.Table.from_batches([], schema=schema)
+        inner = SessionContext().from_arrow(table)
+        return XarrayDataFrame(inner, templates=self._registered_datasets)
+
+    def explain_native(self, query: str) -> str:
```
alxmrs: I don't think this should be a public API

xarray_sql/sql.py
```diff
@@ -174,9 +220,43 @@ def sql(self, query: str, *args, **kwargs) -> XarrayDataFrame:
...
 
+    def _native_sql(self, query: str) -> XarrayDataFrame:
+        """Execute *query* on the native engine and wrap the eager result.
+
+        The native context materialises the result as Arrow batches; we feed
+        them back into a DataFusion ``DataFrame`` (over an in-memory table) so
+        the existing :class:`XarrayDataFrame` round-trip — ``to_pandas`` and
+        ``to_dataset`` — works unchanged.
+        """
+        batches = self._native.sql(query)
```
alxmrs: This needs to be lazy. What if we wanted to load ARCO-ERA5 (several petabytes)

xarray_sql/sql.py
alxmrs: Fixed in a4258a6. The native engine is now lazy: `NativeContext.sql()` returns a lazy `NativeDataFrame` (planned, not executed) that streams result batches instead of collecting, and it round-trips through the same `to_pandas`/`to_dataset` — including the chunked, dask-backed path — as the FFI engine. A reduction or chunked scan over a larger-than-memory store never materialises the whole result. Covered by `test_native_lazy_chunked_roundtrip` and `test_native_sql_returns_lazy_frame`.

---

xarray_sql/sql.py
alxmrs: Fixed in a4258a6 (same lazy-engine change). The result is no longer collected — `NativeDataFrame.execute_stream()` pulls batches on demand (releasing the GIL per batch), and the chunked `to_dataset` path re-plans per chunk with a coordinate filter that pushes into the scan and prunes source partitions. So a `SELECT` over ARCO-ERA5-scale data reads chunk-by-chunk and never holds the whole input or output in memory.

---

xarray_sql/sql.py
alxmrs: Added in ab7a847. The native engine now supports multi-dimension-group datasets as SQL namespaces (`NativeContext.register_table` accepts a schema-qualified name like `era5.surface` and creates the schema on demand) and the `cftime()` filter UDF (a native `ScalarUDFImpl` that calls the Python cftime callable, since a datafusion-python UDF can't cross the FFI boundary). Tests: `test_native_multigroup_namespace` and `test_native_cftime_udf`. Let me know if "this support" meant something else.

---

xarray_sql/sql.py
alxmrs: Agreed — made it private in a4258a6. It's now `_explain_native` (used only by tests to confirm stats/plan shape), so it's no longer part of the public surface.

---

xarray_sql/native.py
```diff
@@ -0,0 +1,104 @@
...
+    __slots__ = ("_batch",)
+
+    def __init__(self, batch: pa.RecordBatch) -> None:
+        self._batch = batch
+
+    def to_pyarrow(self) -> pa.RecordBatch:
+        return self._batch
+
+
+class NativeFrame:
```
alxmrs: Does this serve the entire DataFrame API surface? Can we subclass it so it does?

src/lib.rs
```diff
@@ -631,7 +661,209 @@ impl TableProvider for PrunableStreamingTable {
...
+    stats
+}
+
+/// A thin scan operator that wraps an inner `StreamingTableExec` and reports
+/// exact `Statistics` to the query optimizer.
+///
+/// Execution, schema, ordering, and partitioning are delegated verbatim to the
+/// inner plan (so projection mechanics are reused unchanged); the only thing
+/// this node adds is real cardinality. When consumed natively (not across the
+/// FFI boundary, which drops statistics entirely), this is what lets
```
alxmrs: I think this comment is stale

src/lib.rs
```diff
@@ -631,7 +659,138 @@ impl TableProvider for PrunableStreamingTable {
...
+/// Only `num_rows` is reported — the summed product of each surviving chunk's
+/// dimension sizes, which is exact rather than an estimate. That is what the
+/// cost-based `JoinSelection` rule needs to choose a hash-join build side and
+/// what lets `COUNT(*)` be answered without a scan. Column min/max would add
+/// range hints but do not change those decisions (WHERE filters are already
+/// handled by partition pruning), so they are omitted to keep this simple.
+fn build_scan_statistics(output_schema: &Schema, metas: &[&PartitionMetadata]) -> Statistics {
+    let mut stats = Statistics::new_unknown(output_schema);
+    stats.num_rows = sum_row_counts(metas.iter().copied());
+    stats
```
alxmrs: This may be too simple. I think adding all the relevant stats that we have is good. It may help us later.

src/lib.rs
```diff
@@ -921,12 +1079,24 @@ impl LazyArrowStreamTable {
...
+                            pyo3::exceptions::PyTypeError::new_err(format!(
+                                "each partition must be a (factory, metadata_dict) or \
+                                 (factory, metadata_dict, num_rows) tuple: {e}"
+                            ))
+                        })?;
+                        (f, m, None)
+                    }
+                };
+            let mut meta = convert_python_metadata_from_bound(meta_obj.bind(partitions.py()))?;
+            meta.num_rows = num_rows;
```
alxmrs: Isn't the row count derivable from the metadata dict? If it was, I bet this call could be much simpler.

alxmrs (commented):
Another idea

src/lib.rs
alxmrs: Agreed — expanded in c7550c5. `build_scan_statistics` now reports every stat we can derive from coordinate metadata without touching the data, each exact:

- `num_rows` (as before)
- `total_byte_size` = `num_rows × fixed row width` (derived in Rust from the projected schema's primitive widths; `Absent` if any column is variable-width)
- per dimension-coordinate column: exact `min`/`max` (folded coordinate bounds — the join/filter keys) and `null_count = 0` (grid axes are always fully populated)

Verified through the FFI boundary — `EXPLAIN` on a projected scan now shows `Rows=Exact(2000), Bytes=Exact(48000), [(Col[0]: Min=Exact(Int64(0)) Max=Exact(Int64(3)) Null=Exact(0)), …]`, with data-variable columns correctly carrying no bounds. Pinned by `test_exact_byte_size_in_scan_statistics` and `test_dimension_column_min_max_in_scan_statistics`.

I left `distinct_count` and `sum_value` as `Absent`: distinct count would need per-dimension cardinality (not present in the min/max metadata) plus a coordinate-uniqueness guarantee we don't enforce, and reporting an inexact value as `Exact` would mislead the optimizer.

---

src/lib.rs
alxmrs: Good instinct, but I checked and it doesn't hold in general — the row count isn't recoverable from the metadata dict, for two reasons:

1. **The dict stores min/max, not counts.** A coordinate span of `min=-90.0, max=90.0` (lat at 0.25° spacing) gives no way to know it holds 721 values; datetime spans give a nanosecond range with no step. Only the special case of contiguous unit-step integers would let you back out the count.
2. **Dimensions with unrepresentable bounds are omitted entirely.** `_block_metadata` skips string/object coordinates (and empty slices) — a `station` string dimension simply isn't in the dict, so a product over dict entries would drop that factor and undercount.

Concrete check on `{lat: [-90,0,90], station: ["a","b"]}` (chunk = whole thing):

```
metadata dict: {'lat': (-90.0, 90.0, 'float64')}   # 'station' absent
_block_len:    6                                   # true rows = 3 × 2
```

`_block_len(block)` measures the product of the block's slice extents across **all** dims directly, so it stays correct regardless of coordinate dtype or spacing — which is why it's the source of truth rather than the ranges dict.

The one bit of complexity in this call is tolerating both the 2-tuple `(factory, metadata)` and 3-tuple `(factory, metadata, num_rows)` forms. If you'd rather, I can make `num_rows` a required third element and drop the fallback — that collapses the parse to a single `extract` (I'd update the one 2-tuple test). Kept it optional for now so a source that genuinely doesn't know its row count can still build a table (just without exact stats). Your call.

---

alxmrs: Done in 5f05dce — made `num_rows` a required third tuple element and dropped the fallback. The parse is now a single `item.extract::<(Py, Py, usize)>()`, and since every partition carries an exact count I simplified the downstream too: `PartitionMetadata.num_rows` is a plain `usize` (was `Option`), `sum_row_counts` is a straight fold with no missing-count branch, and `convert_python_ranges_from_bound` just returns the ranges map while `new()` assembles the metadata. Updated the one 2-tuple test and the docstrings/example. Full suite + fmt/clippy/ruff/mypy green.

---

PR #216, opened 2026-07-02
fix: register dimensions without coordinates as columns

### LINKED ISSUE #203: Table that should have N columns only has 2
Repro:

```
  ds = xr.open_dataset(
    's3://carbonplan-share/xbatcher/fashion-mnist-train.zarr',
    engine='zarr',
    chunks={},
    backend_kwargs={'storage_options': {'anon': True}},
  ).reset_coords(drop=True)
# <xarray.Dataset> Size: 189MB
# Dimensions:  (sample: 60000, channel: 1, height: 28, width: 28)
# Coordinates:
#   * sample   (sample) int64 480kB 0 1 2 3 4 5 ... 59995 59996 59997 59998 59999
# Dimensions without coordinates: channel, height, width
# Data variables:
#     images   (sample, channel, height, width) float32 188MB dask.array<chunksize=(7500, 1, 7, 7), meta=np.ndarray>
#     labels   (sample) int64 480kB dask.array<chunksize=(30000,), meta=np.ndarray>

  ctx = xql.XarrayContext()
  ctx.from_dataset(
    "mnist",
    ds,
    chunks=dict(sample=1),
    table_names={("sample", "channel", "height", "width"): "X", ("sample",): "y"}
  )

  ctx.select("select * from mnist.'X' limit 10") # This prints a table with 2 columns, but it should have 5! I think it could be due to how samples is both a coord and a dimension.
```

ghostiee-11:
Dimensions without coordinates (e.g. an image's channel/height/width axes) were dropped from the SQL schema, so `SELECT *` returned too few columns (2 instead of 5 in the #203 repro). This materialises a default integer index for each coordinate-less dimension at the reader boundary, so they show up as columns and keep their absolute position through chunked reads. Adds regression tests. Closes #203.

### Before / after (issue repro)

Before (main), 2 columns:

```
 sample  images
      1     9.0
      1    10.0
      1    11.0
columns: ['sample', 'images']
```

After (this PR), 5 columns, values verified against xarray:

```
 sample  channel  height  width  images
      4        0       0      0    36.0
      4        0       0      1    37.0
      4        0       0      2    38.0
columns: ['sample', 'channel', 'height', 'width', 'images']
values equal xarray to_dataframe(): True
```

Full non-integration test suite passes locally.

### DIFF
--- a/tests/test_df.py
+++ b/tests/test_df.py
@@ -12,6 +12,7 @@ from xarray_sql.df import (
     block_slices,
     compute_chunks,
     dataset_to_record_batch,
+    ensure_default_indexes,
     explode,
     from_map,
     from_map_batched,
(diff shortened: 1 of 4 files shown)


tests/test_sql.py
```diff
@@ -488,3 +488,116 @@ def test_table_names_is_keyword_only(self, mixed_ds):
         ctx = XarrayContext()
         with pytest.raises(TypeError):
             ctx.from_dataset("era5", mixed_ds, {("time",): "x"})
+
+    @pytest.fixture
+    def coordless_dims_ds(self):
+        """Mirror the fashion-mnist layout from issue #203: a dimension
```
alxmrs: Let's not refer to the GH issues in code unless there is a TODO.

tests/test_df.py
```diff
@@ -244,6 +245,84 @@ def test_dataset_to_record_batch_row_count(air_small):
         assert batch.num_rows == expected_rows
 
 
+def _coordless_ds():
+    """A 2-D dataset whose dimensions have NO coordinates (issue #203)."""
```
alxmrs: Let's omit the reference to the GH issue.

alxmrs: Same with the below tests.

tests/test_sql.py
```diff
@@ -488,3 +488,116 @@ def test_table_names_is_keyword_only(self, mixed_ds):
...
+                        n_sample * n_channel * n_height * n_width,
+                        dtype="float32",
+                    ).reshape(n_sample, n_channel, n_height, n_width),
+                ),
+                "labels": (["sample"], np.arange(n_sample, dtype="int64")),
+            },
+            coords={"sample": ("sample", np.arange(n_sample, dtype="int64"))},
+        ).chunk({"sample": 1})
+
+    def test_coordless_dims_appear_as_columns(self, coordless_dims_ds):
```
alxmrs, line 512: I like this test.

xarray_sql/df.py
```diff
@@ -73,6 +73,24 @@ def resolve_chunks(
...
+def ensure_default_indexes(ds: xr.Dataset) -> xr.Dataset:
+    """Attach a default integer index coordinate to every dimension lacking one.
+
+    xarray allows "dimensions without coordinates"; these are absent from
+    ``ds.coords``, so they are dropped from the SQL schema and, once a block is
+    sliced out with ``isel``, their position is synthesized *relative to the
+    block* (restarting at 0 in every partition). Materialising an explicit
+    ``arange`` index up front turns them into ordinary dimension coordinates, so
+    they appear as columns and carry their absolute position through chunked
+    reads (issue #203). Datasets whose dimensions already have coordinates are
```
alxmrs: Let's omit the gh issue ref.

xarray_sql/df.py
```diff
@@ -371,6 +389,11 @@ def iter_record_batches(
 def _parse_schema(ds: xr.Dataset) -> pa.Schema:
     """Extracts a `pa.Schema` from the Dataset, treating dims and data_vars as columns.
 
+    Only *dimension coordinates* become dimension columns, so a dimension
+    without a coordinate would be dropped. Callers must run the Dataset through
+    :func:`ensure_default_indexes` first (the readers do) so every dimension has
+    a coordinate and appears as a column (issue #203).
```
alxmrs: This is a good contract to document. Though, let's omit the issue ref.

xarray_sql/reader.py
```diff
@@ -84,6 +85,7 @@ def __init__(
                 each block dict just before it's converted to Arrow. This
                 allows tests to track when iteration actually occurs.
         """
+        ds = ensure_default_indexes(ds)
```
alxmrs: I think putting this in read_xarray below is better than the constructor.

tests/test_df.py
```diff
@@ -244,6 +245,84 @@ def test_dataset_to_record_batch_row_count(air_small):
         assert batch.num_rows == expected_rows
 
 
+def _coordless_ds():
```
alxmrs: I think we can omit the tests in this file altogether -- the public contract is covered by the sql tests file. Wdyt?

We don't mark it as such, but I see this new utility as a private function.

alxmrs (commented):
Looking good so far. Some notes. Thanks for the quick fix.

tests/test_sql.py
ghostiee-11: Yeah truee.. will take care of this

alxmrs (approved):
LGTM! Thanks for the fix!

PR #218, opened 2026-07-02
fix: infer surviving dims in to_dataset for aggregations (closes #189)

### LINKED ISSUE #189: Can dims be auto-inferred during to_dataset?
```python
import xarray as xr
import xarray_sql as xql

ds = xr.tutorial.open_dataset('air_temperature')

ctx = xql.XarrayContext()
ctx.from_dataset('air', ds, chunks=dict(time=100))

clim = ctx.sql('''
  SELECT
    "time", AVG("air") AS air
  FROM
    "air"
  GROUP BY
    "time"
''')

clim_ds = clim.to_dataset()
```
To my surprise: `ValueError: dims cannot be inferred: no registered Dataset has all of its dims present in the result columns. Pass dims=[...] explicitly.`

I imagine the value for `group by` would become the dim here?

ghostiee-11:
`to_dataset()` inferred dims only when a registered Dataset's entire dim set was present in the result, so a `GROUP BY` that aggregates dims away (the reporter's `SELECT "time", AVG("air") ... GROUP BY "time"`) raised `dims cannot be inferred` instead of using the surviving group-by dim. Inference now uses the registered dims that survive into the result columns (in data-variable axis order), so such aggregations round-trip on the remaining dim(s). Closes #189.

Also: when several registered Datasets imply the same surviving dims, inference returns them rather than raising "ambiguous" (only genuinely differing dims still raise).

### Before / after

Before (main), `.to_dataset()` on the group-by result:

```
ValueError: dims cannot be inferred: no registered Dataset has all of
its dims present in the result columns. Pass dims=[...] explicitly.
```

After (this PR):

```
<xarray.Dataset>
Dimensions:  (time: 4)
Coordinates:
  * time     (time) int64 0 1 2 3
Data variables:
    air      (time) float64 ...
```

Full non-integration test suite passes locally.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,39 @@
+# AGENTS.md
+
+Guidance for contributors (including AI assistants) working on `xarray-sql`. It
+summarizes recurring maintainer review feedback so changes land clean.
+
+## Documentation and comments
+
+- Keep docstrings and comments self-contained. Do **not** put GitHub issue or PR
+  numbers in docstrings or code comments; a reader should not need the issue
+  tracker to understand the code. Issue references belong in the commit message
+  and PR description (e.g. `Closes #189`), not in the source.
+- Do not reference the review conversation, chat, or "the reporter" in comments.
+  Describe the behavior, not how it came up.
+
+## API surface
+
+- Mark internal helpers private with a leading underscore when they are not part
+  of the public API.
+- Prefer doing setup work in the functional entry point (e.g. `read_xarray`)
+  rather than in a class constructor; keep constructors minimal.
+
+## Tests
+
+- Test the public contract (values, dims, coords, attrs), not internal call
+  counts or private classes, so the suite survives refactors.
+- Avoid redundant tests: if a public-path test already covers a behavior, do not
+  add a second lower-level test for the same thing.
+- Make query results deterministic with `ORDER BY` so assertions do not have to
+  re-sort the output.
+- Do not pass `dims=` to `to_dataset()` when inference already resolves them.
+  Reserve explicit `dims=` / `template=` for genuinely ambiguous cases (multiple
+  registered Datasets, or a test that is specifically exercising those
+  arguments).
+
+## Commits
+
+- Use conventional commit prefixes: `fix:`, `feat:`, `refactor:`, `chore:`,
+  `docs:`, `test:`.
+- Keep imports at the top of the file.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,8 +16,6 @@ summarizes recurring maintainer review feedback so changes land clean.
 
 - Mark internal helpers private with a leading underscore when they are not part
   of the public API.
-- Prefer doing setup work in the functional entry point (e.g. `read_xarray`)
-  rather than in a class constructor; keep constructors minimal.
 
 ## Tests
 
@@ -32,8 +30,8 @@ summarizes recurring maintainer review feedback so changes land clean.
   registered Datasets, or a test that is specifically exercising those
   arguments).
 
-## Commits
+## Imports
 
-- Use conventional commit prefixes: `fix:`, `feat:`, `refactor:`, `chore:`,
-  `docs:`, `test:`.
-- Keep imports at the top of the file.
+- Keep imports at the top of the file. Assume transitive dependencies are safe
+  to import non-locally, rather than deferring imports into functions to avoid
+  a dependency.
--- a/README.md
+++ b/README.md
@@ -47,12 +47,8 @@ clim = ctx.sql('''
   ORDER BY month
 ''')
 
-# Write the SQL result back to an Xarray Dataset. `to_dataset()` infers the
-# dimensions from the registered table's dims that survive the query, so a
-# GROUP BY on a real dimension (e.g. `time`) needs no `dims=`. Here `month`
-# is a derived column, not a registered dim, so name it explicitly. The
-# variable's units are recovered from the registered table. One value per
-# month: air(month).
+# Round-trip the result back to Xarray. `month` is a derived column, so name
+# it as the dimension.
 clim_ds = clim.to_dataset(dims=["month"])
 
 # Plot the annual cycle as a time series.
--- a/README.md
+++ b/README.md
@@ -47,9 +47,12 @@ clim = ctx.sql('''
   ORDER BY month
 ''')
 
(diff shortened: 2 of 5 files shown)


tests/test_ds.py
```diff
@@ -139,6 +139,42 @@ def test_aggregation_drops_dim(air_dataset_small):
     np.testing.assert_allclose(actual, expected)
 
 
+def test_aggregation_infers_dims(air_dataset_small):
+    """#189: to_dataset() infers the surviving GROUP BY dims without dims=."""
```
alxmrs: Please add a Claude.md or Agents.md file to the project in this PR, and add a summary of common points of feedback I've given to you in our reviews. 

For example, we shouldn't include GH issue numbers in docstrings (documentation). These should be self contained and not rely on access to the issue tracker.

tests/test_ds.py
```diff
@@ -139,6 +139,42 @@ def test_aggregation_drops_dim(air_dataset_small):
...
+    ctx = XarrayContext()
+    ctx.from_dataset("air", air_dataset_small)
+
+    # GROUP BY lat, lon: time is aggregated away, lat/lon survive.
+    out = ctx.sql(
+        "SELECT lat, lon, AVG(air) AS air_avg FROM air GROUP BY lat, lon"
+    ).to_dataset()
```
alxmrs, line 152: Nice

tests/test_ds.py
```diff
@@ -139,6 +139,42 @@ def test_aggregation_drops_dim(air_dataset_small):
...
 
 
+def test_aggregation_infers_dims(air_dataset_small):
+    """#189: to_dataset() infers the surviving GROUP BY dims without dims=."""
+    ctx = XarrayContext()
+    ctx.from_dataset("air", air_dataset_small)
+
+    # GROUP BY lat, lon: time is aggregated away, lat/lon survive.
+    out = ctx.sql(
+        "SELECT lat, lon, AVG(air) AS air_avg FROM air GROUP BY lat, lon"
```
alxmrs: Let's order by lat desc so we can get rid of a sort in the assert.

tests/test_ds.py
```diff
@@ -139,6 +139,42 @@ def test_aggregation_drops_dim(air_dataset_small):
...
+        air_dataset_small.compute()
+        .sortby(["lat", "lon"])
+        .mean(dim="time")["air"]
+        .values
+    )
+    np.testing.assert_allclose(
+        out.sortby(["lat", "lon"])["air_avg"].values, expected
+    )
+
+    # The reporter's exact case: GROUP BY the time coordinate.
```
alxmrs: Rm this comment or rephrase it to not refer to the chat conversation

tests/test_ds.py
```diff
@@ -383,14 +419,12 @@ def test_to_dataset_multi_registered_requires_explicit_template(
...
-    """If no registered Dataset's dims fit the result -> clear error."""
+def test_to_dataset_infer_fails_when_no_dim_survives(air_dataset_small):
+    """A global aggregation leaves no registered dim in the result -> clear error."""
     ctx = XarrayContext()
     ctx.from_dataset("air", air_dataset_small)
     with pytest.raises(ValueError, match="dims cannot be inferred"):
-        ctx.sql(
-            "SELECT lat, lon, AVG(air) AS air_avg FROM air GROUP BY lat, lon"
-        ).to_dataset()
+        ctx.sql("SELECT AVG(air) AS air_avg FROM air").to_dataset()
```
alxmrs, line 408: Will you also fix all the other tests that have specified dims when they now no longer need them?

alxmrs (commented):
Some notes, overall it's looking great. Happy to have this feature.

AGENTS.md
```diff
@@ -0,0 +1,39 @@
...
+  tracker to understand the code. Issue references belong in the commit message
+  and PR description (e.g. `Closes #189`), not in the source.
+- Do not reference the review conversation, chat, or "the reporter" in comments.
+  Describe the behavior, not how it came up.
+
+## API surface
+
+- Mark internal helpers private with a leading underscore when they are not part
+  of the public API.
+- Prefer doing setup work in the functional entry point (e.g. `read_xarray`)
```
alxmrs: Let's rm this line, it's too specific.

AGENTS.md
```diff
@@ -0,0 +1,39 @@
...
+- Make query results deterministic with `ORDER BY` so assertions do not have to
+  re-sort the output.
+- Do not pass `dims=` to `to_dataset()` when inference already resolves them.
+  Reserve explicit `dims=` / `template=` for genuinely ambiguous cases (multiple
+  registered Datasets, or a test that is specifically exercising those
+  arguments).
+
+## Commits
+
+- Use conventional commit prefixes: `fix:`, `feat:`, `refactor:`, `chore:`,
```
alxmrs: Personally, I don't think these are essential. I could go either way on this standard.

AGENTS.md
```diff
@@ -0,0 +1,39 @@
...
+  `docs:`, `test:`.
+- Keep imports at the top of the file.
```
alxmrs: Let's add a sentence that says: "assume transitive dependencies are safe to import non-locally" -- I believe this is the reason agents do this.

alxmrs (approved):
Ideas about the agent file. Otherwise, this LGTM! Thanks for the fix!

alxmrs:
One more omission in my review! Let's update the README and other docs given this change.

README.md
```diff
@@ -47,9 +47,12 @@ clim = ctx.sql('''
...
 
-# Write the SQL result back to an Xarray Dataset. `month` is a derived
-# column, so name it as the dimension; the variable's units are recovered
-# from the registered table. The result is one value per month: air(month).
+# Write the SQL result back to an Xarray Dataset. `to_dataset()` infers the
+# dimensions from the registered table's dims that survive the query, so a
+# GROUP BY on a real dimension (e.g. `time`) needs no `dims=`. Here `month`
+# is a derived column, not a registered dim, so name it explicitly. The
+# variable's units are recovered from the registered table. One value per
+# month: air(month).
```
alxmrs: This is too much text for a quickstart IMO. We can explain most of this by example.

alxmrs (commented):
Quick note

alxmrs (approved):
LGTM! You may merge when ready.

ghostiee-11:
Yupp.. I also feel this is good now.. and mergeable. But @alxmrs, I dont have merge access.

PR #222, opened 2026-07-04
test: stabilize flaky streaming-aggregation memory test

ghostiee-11:
`test_streaming_aggregation_does_not_explode` flakes intermittently in CI (seen on Python 3.10 and 3.12 on recent PRs), always at ~124-125 MB against its 123.9 MB threshold (`4 * source`).

Root cause: DataFusion processes partitions concurrently, so the tracemalloc peak scales with the number of in-flight partitions (and therefore the runner's core count). Locally the peak varies roughly 1.2x to 4x the source run to run, so the `4x` threshold sits inside that noise band and fails about 1 run in 3.

Fix: sample the peak a few times and assert on the minimum. The best run reflects the true streaming floor (~1.5x source, stable), while a regression that buffers the whole row set would keep every sample high, so the minimum still catches it. The `4x` threshold is unchanged.

Verified: min-of-3 stayed at or below 1.92x across 25 local stress trials (vs the `4x` threshold), and the test passes.

### DIFF
--- a/tests/test_to_dataset_perf.py
+++ b/tests/test_to_dataset_perf.py
@@ -82,26 +82,35 @@ def test_streaming_aggregation_does_not_explode(air_source):
 
     Reduces 3.86M rows of ``air_temperature`` to ~1.3K group cells. The
     aggregation must stream the source once and emit a tiny result, not
-    buffer the entire row set into memory. Reference observation: peak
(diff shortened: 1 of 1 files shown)


alxmrs (approved):
LGTM! Thanks for deflaking.

ghostiee-11:
This can be merged @alxmrs..

PR #224, opened 2026-07-05
fix: register object-dtype and out-of-ns-range coordinates (#223)

### LINKED ISSUE #223: from_dataset fails on object-dtype variables (cftime times, strings) and out-of-ns-range datetimes
### Summary

`XarrayContext.from_dataset` (via `read_xarray_table` -> `df.py`) fails to register several common NetCDF/xarray datasets, because it derives the Arrow schema and block statistics from numpy dtypes alone:

1. **Object-dtype variables/coordinates are rejected.** `_parse_schema` calls `pa.from_numpy_dtype(var.dtype)`, which raises `ArrowNotImplementedError: Unsupported numpy type 17` for object dtype. That is exactly what a **cftime time coordinate** (any non-standard calendar such as `noleap`/`360_day`) or a **string variable** produces.

2. **Datetimes outside the `datetime64[ns]` range overflow.** `_block_metadata` coerces datetime min/max to nanoseconds with `int(pd.Timestamp(min_val).value)`, so coordinates with early/late dates (paleoclimate records, or anything before ~1678 / after ~2262) raise `OverflowError`, even at `us` resolution.

### Reproduce

Object dtype (string variable shown; a cftime time coordinate produces the same `Unsupported numpy type 17`):

```python
import numpy as np, xarray as xr
ds = xr.Dataset({"label": (["x"], np.array(["a", "b"], dtype=object))}, coords={"x": [1, 2]})
XarrayContext().from_dataset("t", ds, chunks={"x": 2})
# pyarrow.lib.ArrowNotImplementedError: Unsupported numpy type 17
```

Out-of-ns-range datetimes:

```python
import numpy as np, xarray as xr
times = xr.date_range("0001-01-01", periods=3, freq="100YS", use_cftime=True).to_datetimeindex(time_unit="us")
ds = xr.Dataset({"v": (["time"], np.arange(3.0))}, coords={"time": times})
XarrayContext().from_dataset("t", ds, chunks={"time": 2})
# OverflowError: Cannot convert Timestamp to nanoseconds without overflow
```

### Real-world impact

NetCDF/climate files routinely use non-standard calendars, so xarray decodes their time coordinate to cftime (object dtype) and (1) fires immediately. Long or paleoclimate records additionally hit (2). Between them a large class of real `.nc` files cannot be registered.

### Suggested fixes

- **`_parse_schema`**: handle object/datetime dtypes rather than passing them straight to `pa.from_numpy_dtype`. Infer from data (`pa.array(...).type`), map cftime/datetime to a `pa.timestamp`, strings to `pa.string()`.
- **`_block_metadata`**: use the array's native unit instead of forcing ns, e.g. `min_val.asm8.view("i8")` with the matching `timestamp_<unit>` tag (the `OverflowError` message itself suggests `.asm8.view('i8')`).

### Environment

xarray-sql 0.2.1, pyarrow 23.0.1, xarray 2026.2.0, pandas 3.0.1, numpy 2.4.2, Python 3.14.3

Found via HoloViz Lumen's `XArraySQLSource`, where uploading a `.nc` with a non-standard-calendar time coordinate hits this. Repros verified locally.

ghostiee-11:
## Overview

`from_dataset` could not register several common NetCDF/xarray datasets: non-cftime object-dtype variables and coordinates (strings) raised `ArrowNotImplementedError`, and datetime coordinates outside the `datetime64[ns]` range raised `OverflowError`. This fixes both in `_parse_schema` and `_block_metadata`, and revives cftime partition pruning that an object-dtype skip had silently disabled. Fixes #223.

## Before

```python
ds = xr.Dataset({"label": (["x"], np.array(["a", "b"], dtype=object))}, coords={"x": [1, 2]})
XarrayContext().from_dataset("t", ds, chunks={"x": 2})
# pyarrow.lib.ArrowNotImplementedError: Unsupported numpy type 17

times = xr.date_range("0001-01-01", periods=3, freq="100YS", use_cftime=True).to_datetimeindex(time_unit="us")
ds = xr.Dataset({"v": (["time"], np.arange(3.0))}, coords={"time": times})
XarrayContext().from_dataset("t", ds, chunks={"time": 2})
# OverflowError: Cannot convert Timestamp to nanoseconds without overflow
```

## After

```python
# both register and query cleanly
c = XarrayContext(); c.from_dataset("t", ds, chunks={"x": 2})
c.sql("SELECT label FROM t ORDER BY x").to_pandas()["label"].tolist()
# ['a', 'b']

# string coordinates, cftime calendars, and pre-1678 / post-2262 dates all register;
# out-of-int64 datetimes skip pruning instead of raising, so registration always succeeds.
```

### DIFF
--- a/tests/test_df.py
+++ b/tests/test_df.py
@@ -536,7 +536,7 @@ def test_compute_chunks_tuples_sum_to_dim_size():
         assert sum(tup) == ds.sizes[dim]
 
 
-# -- Object-dtype and out-of-ns-range coordinate support (issue #223) -------
+# -- Object-dtype and out-of-ns-range coordinate support --------------------
 
 
 def _field_type(schema, name):
@@ -584,10 +584,9 @@ def test_partition_metadata_skips_out_of_ns_datetime():
     assert all("time" not in m for m in meta)
 
 
-def test_parse_schema_all_null_object_var_defaults_to_string():
-    # An all-null (or empty) object column has no data to infer from; object
-    # dtype in xarray almost always means strings, so default to pa.string()
-    # rather than a null column.
+def test_parse_schema_all_null_object_var_stays_null():
+    # An all-null object column has no data to infer a type from; let null be
+    # null rather than coercing it to a string column.
     ds = _ensure_default_indexes(
         xr.Dataset(
             {"label": (["x"], np.array([None, None], dtype=object))},
@@ -595,7 +594,7 @@ def test_parse_schema_all_null_object_var_defaults_to_string():
         )
     )
     schema = _parse_schema(ds)
-    assert schema.field("label").type == pa.string()
+    assert pa.types.is_null(schema.field("label").type)
 
 
 def test_partition_metadata_prunes_cftime_coord():
--- a/tests/test_df.py
+++ b/tests/test_df.py
@@ -8,6 +8,7 @@ import xarray as xr
 
 from xarray_sql.df import (
     DEFAULT_BATCH_SIZE,
+    _ensure_default_indexes,
     _parse_schema,
     block_slices,
     compute_chunks,
@@ -16,6 +17,7 @@ from xarray_sql.df import (
     from_map,
     from_map_batched,
     iter_record_batches,
+    partition_metadata,
     pivot,
 )
 from xarray_sql.reader import read_xarray, read_xarray_table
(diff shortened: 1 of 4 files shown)


tests/test_df.py
```diff
@@ -532,3 +534,139 @@ def test_compute_chunks_tuples_sum_to_dim_size():
     result = compute_chunks(ds, {"a": 3, "b": 4, "c": 5})
     for dim, tup in result.items():
         assert sum(tup) == ds.sizes[dim]
+
+
+# -- Object-dtype and out-of-ns-range coordinate support (issue #223) -------
```
alxmrs: Let's omit this GH issue.

alxmrs: Did you self review this PR?

tests/test_df.py
```diff
@@ -532,3 +534,139 @@ def test_compute_chunks_tuples_sum_to_dim_size():
...
+
+def _field_type(schema, name):
+    return schema.field(name).type
+
+
+def test_parse_schema_maps_object_string_data_var_to_string():
+    # A string variable arrives as numpy object dtype; _parse_schema must not
+    # hand it to pa.from_numpy_dtype (which raises "Unsupported numpy type 17").
+    ds = xr.Dataset(
+        {"label": (["x"], np.array(["a", "b"], dtype=object))},
```
alxmrs, line 550: Are there types of objects that don't map to strings? Do we throw errors for these cases, or just try to convert them to pa.string()?

tests/test_df.py
```diff
@@ -532,3 +534,139 @@ def test_compute_chunks_tuples_sum_to_dim_size():
...
+    # A string dimension coordinate (e.g. station names) is object dtype too.
+    ds = xr.Dataset(
+        {"v": (["station"], [1.0, 2.0])},
+        coords={"station": np.array(["A", "B"], dtype=object)},
+    )
+    schema = _parse_schema(_ensure_default_indexes(ds))
+    assert _field_type(schema, "station") == pa.string()
+
+
+def test_partition_metadata_skips_out_of_ns_datetime():
```
alxmrs, line 567: What are the consequences of skipping this for cftime typed time ranges? What are the tradeoffs?

tests/test_df.py
```diff
@@ -532,3 +534,139 @@ def test_compute_chunks_tuples_sum_to_dim_size():
...
+    blocks = list(block_slices(ds, chunks={"time": 2}))
+
+    meta = partition_metadata(ds, blocks)  # must not raise
+
+    assert len(meta) == len(blocks)
+    # "time" is unpruneable here, so it is omitted from every partition.
+    assert all("time" not in m for m in meta)
+
+
+def test_parse_schema_all_null_object_var_defaults_to_string():
```
alxmrs: This seems like strictly incorrect behavior. Let's let null be null.

tests/test_df.py
```diff
@@ -532,3 +534,139 @@ def test_compute_chunks_tuples_sum_to_dim_size():
...
+
+    meta = partition_metadata(ds, blocks)
+
+    assert all("time" in m for m in meta)
+    for m in meta:
+        _, _, tag = m["time"]
+        assert tag == "timestamp_ns"
+
+
+def test_partition_metadata_skips_ancient_cftime():
```
alxmrs, line 619: Does pa have some sort of bigint type that we could use to handle these cftime cases instead of skipping pruning?

alxmrs: I can't tell yet what the better approach is (it's probably the implemented one).

tests/test_df.py
```diff
@@ -532,3 +534,139 @@ def test_compute_chunks_tuples_sum_to_dim_size():
...
+    assert batch.column("label").to_pylist() == ["a", "b", "c", "d"]
+
+    # The streaming path must agree with the one-shot path.
+    streamed = pa.Table.from_batches(
+        list(iter_record_batches(ds, schema, batch_size=2)), schema=schema
+    )
+    assert streamed.column("label").to_pylist() == ["a", "b", "c", "d"]
+
+
+def test_partition_metadata_in_range_datetime_still_pruned():
```
alxmrs, line 657: Happy to have this test.

xarray_sql/cftime.py
```diff
@@ -170,20 +170,34 @@ def convert_for_field(values, field: pa.Field) -> np.ndarray:
 # ---------------------------------------------------------------------------
 
 
+#: Bounds of a signed 64-bit integer; a pruning bound outside this range
+#: cannot be handed to the Rust ``ScalarBound`` layer.
+_INT64_MIN: int = -(2**63)
+_INT64_MAX: int = 2**63 - 1
```
alxmrs: Can we get this from a more authoritative place, like a numpy or pa type? Or a python std library? This is the right calculation I'm sure, but I'd rather use something more maintainable.

xarray_sql/df.py
```diff
@@ -386,6 +386,31 @@ def iter_record_batches(
...
+    what a string variable or coordinate produces. Object-dtype arrays are
+    never Dask/Zarr-backed, so letting Arrow infer the type from the data with
+    ``pa.array`` (e.g. ``pa.string()`` for Python strings) is safe and triggers
+    no remote I/O.
+
+    Only the first ``sample_size`` elements are inspected so a huge in-memory
+    object column is not fully copied just to read its type. An empty or
+    all-null sample yields ``pa.null()``, which is meaningless as a column
+    type; object dtype in xarray almost always means strings, so fall back to
+    ``pa.string()`` in that case.
```
alxmrs: Let's be correct, not polite

xarray_sql/df.py
```diff
@@ -386,6 +386,31 @@ def iter_record_batches(
...
+    """
+    sample = np.asarray(values).ravel()[:sample_size]
+    arrow_type = pa.array(sample).type
+    if pa.types.is_null(arrow_type):
+        return pa.string()
+    return arrow_type
```
alxmrs: I recommend reviewing the xr and pa sources to better understand what object types could mean.

xarray_sql/df.py
```diff
@@ -422,11 +451,18 @@ def _parse_schema(ds: xr.Dataset) -> pa.Schema:
...
         # cheap guard.  Only fall back to _is_cftime (which materializes
         # element 0) when dtype is object.
-        if var.dtype == np.dtype("O") and cft.is_cftime(var.values):
-            # Rare: a data variable holding cftime objects.  Use same encoding
-            # as the first cftime dimension coordinate, or default.
-            cal = var.values.ravel()[0].calendar
-            columns.append(cft.arrow_field(var_name, cft.DEFAULT_UNITS, cal))
+        if var.dtype == np.dtype("O"):
+            if cft.is_cftime(var.values):
+                # Rare: a data variable holding cftime objects.  Use same
```
alxmrs: I don't think this is that rare.

alxmrs (changes requested):
Some questions and feedback for review.

tests/test_df.py
ghostiee-11: checked the pa + xr behaviour. i don't force pa.string() anymore, pa.array infers per content: strings->string, bytes->binary, ints->int64, datetime->timestamp[us]. all-null stays null. a column mixing incompatible types (e.g. str+int) raises ArrowTypeError, so we surface a clear error instead of silently coercing.

tests/test_df.py
ghostiee-11: skipping only drops partition pruning (a query-time optimization) for that dim, correctness is unaffected. the rust pruner treats a missing dim as "never prune", so filtered queries just scan all partitions. only bites out-of-ns-range dates (pre-1678 / post-2262), which are rare.

tests/test_df.py
ghostiee-11: had a look. the rust ScalarBound only takes timestamp_ns / int64 / float64, and bound_to_scalar only maps a TimestampNanos bound onto a Timestamp column. an int64 bound wouldn't apply to a timestamp[us] column so it can't prune it. representing the column as int64 micros would prune but breaks string-timestamp SQL filters. so skipping pruning for those rare out-of-range dates is the safe call, went with the implemented approach.

alxmrs: Had a check around. It looks like this is related to this issue: https://github.com/apache/arrow/issues/32664. 

Pyarrow doesn't support larger than int64, but arrow does: https://arrow.apache.org/rust/arrow/datatypes/struct.i256.html

I think we file an issue associated with this linked to the pyarrow issue. We can see how much this affects users, and if it does, one workaround we can take it to bridge from cftime types to large pyarrow types in our own native sources (complicated, so we can delay based off demand).

alxmrs (approved):
LGTM! I'll merge this soon. I have one comment to create a follow up issue later on. 

Thanks for this contribution, Aman.

PR #226, opened 2026-07-11
Introduced new test_read_netcdf.py function for directly converting netCDF files into XarrayContext.

dfang843:


xarray_sql/sql.py
```diff
@@ -156,6 +157,55 @@ def _maybe_register_cftime_udf(self, ds: xr.Dataset) -> None:
                     self.register_udf(cft.make_cftime_udf(units, cal))
                     break  # One UDF per context is enough.
 
+    @classmethod
```
alxmrs, line 160: I think this should be an instance method.

xarray_sql/sql.py
```diff
@@ -156,6 +157,55 @@ def _maybe_register_cftime_udf(self, ds: xr.Dataset) -> None:
...
+            Additional kwargs forwarded to xr.open_dataset.
+
+        Returns
+        -------
+        XarrayContext
+            A context with the dataset already registered.
+        """
+        path = Path(path)
+
+        ds = xr.open_dataset(path, engine=engine, **open_kwargs)
```
alxmrs, line 194: The engine should be best-effort a netcdf engine IMO, and users won't need to pass it in, since the method name implies the type of data.

xarray_sql/sql.py
```diff
@@ -156,6 +157,55 @@ def _maybe_register_cftime_udf(self, ds: xr.Dataset) -> None:
...
+            xarray backend engine. Default is 'netcdf4'.
+        **open_kwargs
+            Additional kwargs forwarded to xr.open_dataset.
+
+        Returns
+        -------
+        XarrayContext
+            A context with the dataset already registered.
+        """
+        path = Path(path)
```
alxmrs, line 192: I think open dataset handles this.

xarray_sql/sql.py
```diff
@@ -156,6 +157,55 @@ def _maybe_register_cftime_udf(self, ds: xr.Dataset) -> None:
...
+    @classmethod
+    def read_netcdf(
+      cls,
+      path: str | Path,
+      table_name: str | None = None,
+      chunks: dict | None = None,
+      engine: str | None = "netcdf4",
+      **open_kwargs,
+    ):
+        """
```
alxmrs, line 169: We use Google style docstrings.

xarray_sql/sql.py
```diff
@@ -156,6 +157,55 @@ def _maybe_register_cftime_udf(self, ds: xr.Dataset) -> None:
...
+        ds = xr.open_dataset(path, engine=engine, **open_kwargs)
+
+        if not isinstance(ds, xr.Dataset):
+            raise TypeError(f"Expected xr.Dataset, got {type(ds).__name__}")
+        
+        if table_name is None:
+            table_name = path.stem
+
+        if chunks is None:
+            chunks = {"time": 24} if "time" in ds.dims else {d: -1 for d in ds.dims}
```
alxmrs, line 203: Let's be correct, not polite. The users must pass in chunks.

xarray_sql/sql.py
```diff
@@ -7,7 +7,8 @@
 from .df import Chunks
 from .ds import XarrayDataFrame
 from .reader import read_xarray_table
-
+from pathlib import Path
+from matplotlib.dates import date2num
```
alxmrs, line 11: Why do we import this?

test_read_netcdf.py
```diff
@@ -0,0 +1,11 @@
+from metpy.cbook import get_test_data
+from xarray_sql import XarrayContext
+
+path = get_test_data('irma_gfs_example.nc', False)
```
alxmrs, line 4: Let's use pytest conventions for tests.

alxmrs (changes requested):
First (quick) pass. Thanks for the PR!

xarray_sql/sql.py
dfang843: These libraries have been imported to ensure that test_read_netcdf.py doesn't throw errors when running.

PR #230, opened 2026-07-23
fix: projection dropping a cftime coord crashes batch reading

ghostiee-11:
A projection that drops a cftime dimension coordinate (for example `GROUP BY` on a non-time dim, which removes `time` from the scan) crashed with `KeyError: 'Column time does not exist in schema'` during batch reading.

`iter_record_batches` preloaded every dimension's coord values and called `schema.field(name)` for cftime coords, but the projected batch schema no longer contains the dropped dim. datetime64 coords skip that branch, so only cftime datasets broke.

The dropped dim is never read in the batch loop (it iterates only the schema's fields), so this skips preloading any dim absent from the projected schema.

Verified across all calendars (Gregorian/NoLeap mapped to timestamp, 360-day/Julian mapped to int64) and every projected dimension. The regression test in `test_df.py` fails without the fix (`KeyError`) and passes with it.

### DIFF
--- a/tests/test_df.py
+++ b/tests/test_df.py
@@ -181,6 +181,30 @@ def test_iter_record_batches_matches_dataset_to_record_batch(air_small):
     pd.testing.assert_frame_equal(actual_df, expected_df)
 
 
+def test_iter_record_batches_projection_drops_cftime_dim():
+    """A projection that drops a cftime dim (e.g. time under GROUP BY level)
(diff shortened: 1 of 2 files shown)


ghostiee-11:
@alxmrs, Can you please review this

alxmrs:
Sorry for the late reply. TAL now.

alxmrs (approved):
LGTM! Thanks for the fix.

PR #245, opened 2026-08-10
Replace searchsorted with a hash index in the reverse-pivot scatter

Mmoncadaisla:
## Summary

Speed up converting SQL results back into xarray arrays by using a reusable hash index to locate values on irregular coordinate axes.

Reconstruction benchmarks show a **2.1–3.3× speedup** for the tested irregular-axis workloads. Uniformly spaced axes retain their existing arithmetic shortcut.

The tradeoff is additional temporary memory for the hash index.

## Why this helps

SQL engines return rows, while xarray stores values in multidimensional arrays. Reconstructing a Dataset requires finding the correct array position for each row’s coordinates.

For example, given a requested station axis:

    station = [10, 25, 40, 100]

Rows arriving in this order:

    station      temperature
    40           18.5
    10           21.0
    100          16.2
    25           20.3

Must be placed at positions:

    [2, 0, 3, 1]

Producing:

    temperature = [21.0, 20.3, 18.5, 16.2]

Previously, irregular axes used a sorted copy of the coordinates and a binary search for each incoming value. With millions of rows and several dimensions, those repeated searches can become a reconstruction bottleneck.

This PR builds a pandas index once per dimension for each output-array reconstruction and reuses it across batches.

## Implementation

A private `_CoordLookup` helper selects the appropriate strategy:

| Coordinate axis | Lookup strategy |
|---|---|
| Uniformly spaced | Existing arithmetic calculation |
| Irregular with unique values | New hash lookup using `pd.Index.get_indexer` |
| Duplicate values or a dtype unsupported by pandas indexing | Existing `argsort` + `searchsorted` fallback |

For irregular unique axes, the lookup changes from approximately O(log n) per value to O(1) average time after index construction.

The helper is used by the shared scatter operation, covering both eager reconstruction and engine-backed lazy reads. Coordinate ordering and duplicate-coordinate fallback behavior are preserved.

## Performance

Measured through `xql.to_dataset` using shuffled rows over an irregular axis on a GCP `n2-standard-16` machine. Results are medians of five runs.

| Cells | Previous lookup | Hash lookup | Speedup |
|---|---:|---:|---:|
| 24 K | 4.0 ms | 1.9 ms | 2.1× |
| 240 K | 40.4 ms | 15.2 ms | 2.7× |
| 2.4 M | 481 ms | 162 ms | 3.0× |
| 6 M | 1,945 ms | 624 ms | 3.1× |
| 24 M | 13,685 ms | 4,168 ms | 3.3× |

These measurements cover **reconstruction**, not total SQL query execution time.

Ordered and unordered ERA5 reconstruction benchmarks remained approximately unchanged across DataFusion, DuckDB, and Polars.

## Memory tradeoff

On a two-million-value float64 axis, the hash lookup used approximately **34 bytes per coordinate value**, compared with **16 bytes** for the previous sorted lookup structures.

This memory is temporary and associated with each non-affine dimension’s lookup. Uniformly spaced axes allocate neither structure.

## Error handling

The new hash path raises a `ValueError` when an incoming coordinate is absent from the requested output axis. The message identifies the dimension and includes an example missing value.

Previously, such a mismatch could cause an incorrect write or a less informative failure.

This validation applies to irregular unique axes using the hash path. The arithmetic and sorted-search paths retain their existing behavior.

## Validation

- 96 focused tests passed, covering Dataset reconstruction, lazy round-trips, DuckDB integration, and the new coordinate lookup tests.
- Additional multi-batch checks passed for irregular floats, descending uniformly spaced coordinates, float16, strings, NaN/NaT, timedeltas, large signed and unsigned integers, and differing datetime units.
- CI checks are passing.

The new test module covers irregular-axis reconstruction, NaN coordinates, missing-coordinate errors, and duplicate-axis fallback. Its descending-axis and float16 cases currently take the existing reshape shortcut; those fixtures should be strengthened to exercise their intended lookup strategies directly.

### DIFF
--- a/xarray_sql/ds.py
+++ b/xarray_sql/ds.py
@@ -198,14 +198,14 @@ class _CoordLookup:
       batches. The table holds roughly twice the transient memory of
       the sorted copy the search strategy needs; affine axes build
       neither.
-    * **Sorted search** — axes with duplicate values, where a unique-key
-      hash table cannot represent the value-to-position mapping:
+    * **Sorted search** — axes the hash strategy cannot represent:
+      duplicate values (a unique-key table has no single position per
+      value) or dtypes ``pd.Index`` does not support (e.g. float16):
       ``np.argsort`` once, then ``np.searchsorted`` per batch (O(log n)
-      per row). Duplicate dim values only reach this code from a
-      pathological result (``to_dataset`` raises on duplicate dim tuples
-      earlier on the main paths), but the fallback keeps the mapping
-      well-defined: each value resolves to one of the positions holding
-      it (which one is unspecified).
+      per row). With duplicates, each value resolves to one of the
+      positions holding it (which one is unspecified), and rows
+      targeting the same cell overwrite in batch order — the scatter's
+      last-write-wins semantics.
 
     Only the hash strategy can detect a probe value absent from the
     axis; ``positions_for`` then raises ``ValueError`` — a symptom of a
@@ -222,8 +222,15 @@ class _CoordLookup:
         self._sorted_idx: np.ndarray | None = None
         self._sorted_req: np.ndarray | None = None
         if self._affine is None:
-            index = pd.Index(requested)
-            if index.is_unique:
+            index: pd.Index | None
+            try:
+                index = pd.Index(requested)
+            except (NotImplementedError, TypeError):
+                # Dtypes pandas cannot index (e.g. float16) take the
+                # sorted-search strategy, which only needs numpy
+                # comparisons.
+                index = None
+            if index is not None and index.is_unique:
                 self._hash_index = index
             else:
                 self._sorted_idx = np.argsort(requested)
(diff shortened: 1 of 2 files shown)


xarray_sql/ds.py
```diff
@@ -172,12 +172,92 @@ def _affine_axis(requested: np.ndarray) -> tuple[float, float] | None:
...
+                # sorted-search strategy, which only needs numpy
+                # comparisons.
+                index = None
+            if index is not None and index.is_unique:
+                self._hash_index = index
+            else:
+                self._sorted_idx = np.argsort(requested)
+                self._sorted_req = requested[self._sorted_idx]
+
+    def positions_for(self, vals: np.ndarray, dim: str) -> np.ndarray:
```
alxmrs, line 239: This looks slick at the call site. Very well done.

alxmrs (approved):
Solid PR. I like the abstracts you add here and the performance results speak for themselves.

PR #246, opened 2026-08-19
Versioned documentation: keep the public default on the latest stable release

Mmoncadaisla:
## Why

Publishing the `v0.4.0-rc.1` pre-release replaced the public docs site with unreleased content: the site deploys from `main` as a single version, and `main` now *is* the RC. Pre-release docs should be available, but the public default should stay on the latest stable release.

## What

Docs are now versioned with the [Zensical fork of mike](https://zensical.org/docs/setup/versioning/). Each version lives in its own directory on a `gh-pages` branch, with a version selector on every page. Pages still publishes through the existing artifact flow, so the Pages source setting stays as it is.

| Event | Deploys to | `latest` (public default) |
|---|---|---|
| Push to `main` | `dev/` | unchanged |
| Pre-release published | rolling `rc/`, titled with the tag | unchanged |
| Stable release published | `X.Y/` | moves here |
| Manual run | any version, built from any tag | only if requested |

Safeguards worth knowing about:

- `latest` never moves backward (a `v0.3.4` after `v0.4.0` won't repoint the site) and can never point at `rc`.
- A `-rc.N` suffix in the tag routes to `rc/` even if the pre-release checkbox was forgotten.
- Rolling versions (`dev`, `rc`) are newest-wins under concurrent builds; stable releases deploy in publish order.
- Release tags are validated before use, and the mike fork is pinned to a commit (it installs from a mutable branch in a job with `contents: write`).

## After merging (one-time)

1. **Allow tag deploys**: Settings → Environments → `github-pages` → deployment branches and tags → add tag rule `v*`. Today only `main` may deploy, which would reject release-triggered runs.
2. **Backfill the stable docs**: Actions → "Build documentation" → Run workflow on `main` with ref `v0.3.3`, version `0.3`, "latest" ticked.

Until step 2 runs, the old site stays live — the workflow refuses to publish a site that has no default version yet.

## Notes

- Docs deploy independently of the PyPI publish. If a release goes bad, CONTRIBUTING now documents how to move `latest` back and delete a version.
- Verified locally end to end: version deploys, alias moves, root redirect, `rc` titling, idempotent re-runs, `zensical build`, and the unit suite.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,7 +15,7 @@ on:
         type: string
         default: ""
       version:
-        description: "Docs version to deploy (e.g. 0.3)"
+        description: "Docs version to deploy (e.g. 0.3, or rc for the rolling pre-release docs)"
         required: true
         type: string
       latest:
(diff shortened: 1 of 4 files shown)


alxmrs (approved):
LGTM, thank you for finding the right set of documentation generating technologies.

PR #227, opened 2026-07-14
Multi-engine backends: pushdown DuckDB adapter, engine-neutral Arrow dataset, caching helpers

Mmoncadaisla:
## Summary

This branch prototypes **xarray-sql as the Xarray ↔ engine translator**: the library owns exactly two seams — **register** (a lazy `xarray.Dataset` becomes a table on an engine's own connection) and **round-trip** (any engine's Arrow result becomes a labeled `xr.Dataset` again). SQL dialects, geometry functions, H3, and optimizers stay with each engine and its extension ecosystem. No transpiler, no unified dialect.

```mermaid
flowchart TB
    A["Zarr · NetCDF · GRIB · GeoTIFF · Earth Engine"] --> B["lazy xarray.Dataset"]
    B --> C["seam 1 — register<br/>xql.register(con, name, ds) / xql.arrow_dataset(ds)"]
    C --> D["DuckDB<br/>+ spatial, h3, …"]
    C --> E["Polars"]
    C --> F["DataFusion"]
    C --> G["Dask"]
    D --> H["seam 2 — round-trip<br/>xql.to_dataset(result, template=ds)"]
    E --> H
    F --> H
    H --> I["labeled xr.Dataset — SQL in, array out"]
```

One object serves every engine: `xql.arrow_dataset(ds)` is a real `pyarrow.dataset.Dataset` (the pattern Lance uses for `LanceDataset`), so DuckDB registers it, Polars scans it via `scan_pyarrow_dataset`, DataFusion consumes it via `register_dataset` (through the fragments API), and Dask maps over `get_fragments()`. Ibis works through `ibis.duckdb.from_connection` with zero code.

## How the pushdown scan works

```mermaid
flowchart TB
    Q["engine calls scanner(columns, filter)"] --> P["prune chunks<br/>per-dimension shadow fragments;<br/>Arrow guarantee simplification decides satisfiability"]
    Q --> J["project<br/>only referenced variables are read"]
    P --> C["coalesce (opt-in)<br/>merge consecutive surviving chunks<br/>into single reads"]
    C --> L["prefetch pool<br/>bounded by prefetch (threads)<br/>and prefetch_bytes (memory)"]
    J --> L
    L --> X["exact filter<br/>pyarrow applies the pushed expression row-exactly"]
    X --> R["Arrow batches → engine"]
```

Three properties worth calling out:

[...]

PR #252, opened 2026-09-05
🤖 Name your tables on every engine: table_names across the multi-engine backends

alxmrs:
This feature adds the `table_names` feature to all engine backends so each SQL engine behaves similar to the native datafusion experience.

Disclaimer: more slop involved in the process of this PR because I had a baby and otherwise have my hands full. 

🤖  below (with human edits).

> 🤖 This PR was drafted by Claude Code (Opus 5) and is up for human review.

```python
ctx.from_dataset('era5', ds, chunks=dict(time=6), table_names={
    ('time', 'latitude', 'longitude'): 'surface',
    ('time', 'level', 'latitude', 'longitude'): 'atmosphere',
})

ctx.sql('SELECT AVG("2m_temperature") FROM era5.surface WHERE ...')
```

The `table_names` keyword only ever reached DataFusion. Since #227 a Dataset can be
registered on DuckDB or scanned by Polars, and on those paths the tables went
back to being named after their dimensions. This routes `table_names` through
the register seam, so it means the same thing wherever a Dataset is
registered:

```python
con = duckdb.connect()
xql.register(con, 'era5', ds, table_names={
    ('time', 'latitude', 'longitude'): 'surface',
    ('time', 'level', 'latitude', 'longitude'): 'atmosphere',
})

con.sql('SELECT AVG("2m_temperature") FROM era5.surface WHERE ...')
```

## What each engine does

**DuckDB.** `con.register` can only place an object in DuckDB's temporary
namespace, so each dimension group is still registered flat as `era5_surface`,
and the groups are additionally mirrored as views in a schema of their own so
`era5.surface` resolves. Both spellings hit the same scan — chunk pruning and
column projection travel through the view (pinned by a test: a `WHERE time =
...` query reads 1 of 6 chunks and projects two columns). A read-only
connection cannot create a schema; that path warns and leaves the flat tables
rather than losing the registration.

**A plain `datafusion.SessionContext`.** It now splits mixed-dimension
Datasets into a schema the way `XarrayContext` does. Previously it registered
the whole Dataset as a single table, which a mixed-dimension Dataset cannot
be.

**Polars**, and anything else consuming the pyarrow dataset protocol, has no
connection object to dispatch on, so it takes the tables directly:

```python
tables = xql.arrow_datasets(ds, 'era5', table_names={...})

ctx = pl.SQLContext()
for table, dataset in tables.items():   # 'era5_surface', 'era5_atmosphere'
    ctx.register(table, pl.scan_pyarrow_dataset(dataset))
```

`arrow_datasets` is the new public function here — `arrow_dataset` wants a
Dataset whose variables share one set of dimensions, and this is the
mixed-dimension form of it. It reads the shared dimension coordinates once for
all the tables it returns, which is a network round-trip saved per dimension
per group on Zarr-backed stores.

## Naming rules

One rule, `df.resolve_table_names`, serves every adapter:

- Groups you do not name keep their dimensions joined by underscores; the
  group holding scalar variables, if any, is `scalar`.
- Keys naming a dimension group the Dataset does not have are ignored, so one
  canonical naming map can be reused across Datasets holding different subsets
  of the same variables.
- Two groups claiming the same name raises, instead of silently registering
  one table over the other.

## Also here

The geospatial benchmark suite carried its own copy of the split
(`_group_tables`), written when the library had no cross-engine version. It
now calls `arrow_datasets`, so benchmark naming cannot drift from the
library's.

`docs/limitations.md` listed "mixed-dimension datasets split into one DuckDB
table per dim group" as a sharp edge; it is now the narrower read-only-
connection caveat.

## Testing

`tests/test_table_names.py` holds the contract in one place: the same naming
map on `XarrayContext`, a plain `SessionContext`, DuckDB, and Polars,
including a parity test that runs one SQL string against `era5.surface` on
three engines and asserts one number. Plus the spellings (dotted and flat),
pushdown through the view, the round-trip back to a Dataset, the read-only
warning, and the two naming rules above.

Full suite: 332 passed, 2 skipped. `ruff` and `mypy` clean.

The benchmark cases themselves need ARCO-ERA5 from GCS, so the harness change
was exercised against a synthetic ERA5-shaped Dataset on all three engines
rather than through a suite run.

## Not done

`CHANGELOG.md` is untouched — its `Unreleased` section predates both 0.4.0-rc.1
and #227, so it looked retired. Happy to add an entry if it is still live.

https://claude.ai/code/session_013LfwhbX9VLmvoN6ka7py5w

### DIFF
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -144,12 +144,14 @@ A materialized table or bare C-stream has no query behind it, so the
 re-execution form of `chunks=` cannot serve it — `spill=True` (one
 pass to a temporary Parquet file) is the chunked path for these.
 
-### Mixed-dimension datasets split into one DuckDB table per dim group
+### A read-only DuckDB connection loses the dotted table names
 
-DuckDB registration has no schema namespace, so variables with
-different dims land in suffixed tables (`<name>_<dims>`), sharing one
-set of coordinate reads. DataFusion registers the same layout as
-`name.group` tables inside one schema.
+Mixed-dimension Datasets split into one table per dimension group on
+every engine, addressed as `name.group`. On DuckDB that dotted spelling
+is a view in a schema, because `con.register` reaches only the
+temporary namespace — and creating a schema needs a writable catalog.
+Registering on a read-only connection therefore warns and leaves the
+flat `name_group` tables, which are always registered and always work.
 
 ### Pointwise indexers on lazy round-trip arrays are slower
 
--- a/xarray_sql/backends/base.py
+++ b/xarray_sql/backends/base.py
@@ -42,13 +42,7 @@ class EngineAdapter(Protocol[ConT]):
         table_names: TableNames = None,
         **kwargs: Any,
     ) -> ConT:
-        """Register *ds* as table *name* on *con*; returns *con*.
-
-        ``table_names`` names the per-dimension-group tables a
-        mixed-dimension Dataset splits into (see
-        [resolve_table_names][xarray_sql.df.resolve_table_names]);
-        adapters that split must honour it.
-        """
+        """Register *ds* as table *name* on *con*; returns *con*."""
         ...
 
 
--- a/xarray_sql/backends/base.py
+++ b/xarray_sql/backends/base.py
@@ -18,7 +18,7 @@ from typing import Any, Protocol, TypeGuard, TypeVar, cast
 
 import xarray as xr
 
-from ..df import Chunks
+from ..df import Chunks, TableNames
 
 ConT = TypeVar("ConT")
 """An engine's native connection type (e.g. ``duckdb.DuckDBPyConnection``)."""
@@ -39,9 +39,16 @@ class EngineAdapter(Protocol[ConT]):
         ds: xr.Dataset,
         *,
         chunks: Chunks = None,
+        table_names: TableNames = None,
         **kwargs: Any,
     ) -> ConT:
-        """Register *ds* as table *name* on *con*; returns *con*."""
+        """Register *ds* as table *name* on *con*; returns *con*.
+
+        ``table_names`` names the per-dimension-group tables a
+        mixed-dimension Dataset splits into (see
+        [resolve_table_names][xarray_sql.df.resolve_table_names]);
+        adapters that split must honour it.
+        """
         ...
 
 
@@ -74,6 +81,7 @@ def register(
     ds: xr.Dataset,
     *,
     chunks: Chunks = None,
+    table_names: TableNames = None,
     **kwargs: Any,
 ) -> ConT:
     """Register a lazy xarray Dataset as a table on an engine connection.
@@ -94,20 +102,37 @@ def register(
         rel = con.sql("SELECT time, AVG(t2m) AS t2m FROM era5 GROUP BY time")
         result = xql.to_dataset(rel, template=ds)
 
+    A Dataset whose variables sit on different dimensions is split into
+    one table per dimension group. Name those tables with
+    ``table_names``, and the same SQL runs on every engine::
+
+        xql.register(con, "era5", ds, table_names={
+            ("time", "latitude", "longitude"): "surface",
+            ("time", "level", "latitude", "longitude"): "atmosphere",
+        })
+        con.sql("SELECT AVG(temperature) FROM era5.atmosphere")
+
     Args:
         con: An engine connection: a ``datafusion.SessionContext`` (or
             [xarray_sql.XarrayContext][]) or a
             ``duckdb.DuckDBPyConnection``.
         name: The table name to register the Dataset under. Datasets
             whose variables have differing dimensions are split into one
-            table per dimension group (a SQL schema ``name.group`` on
-            DataFusion; ``name_group`` tables on DuckDB).
+            table per dimension group, addressed as ``name.group`` on
+            every engine (DuckDB also keeps the flat ``name_group``
+            spelling, since its registration namespace is flat).
         ds: An xarray Dataset.
         chunks: Xarray-like chunks specification controlling partition
             granularity. Defaults to the Dataset's existing chunks.
+        table_names: Maps a dimension group's exact dim tuple to the name
+            its table takes. Groups left unnamed take their dimensions
+            joined by underscores (``time_latitude_longitude``); the
+            group holding scalar variables, if any, takes ``scalar``.
+            Keys matching no group in ``ds`` are ignored, so one naming
+            map can be reused across Datasets holding different subsets
+            of the same variables.
         **kwargs: Adapter-specific options, forwarded as-is — e.g.
-            ``table_names`` on DataFusion, ``batch_size`` / ``prefetch``
-            on DuckDB.
+            ``batch_size`` / ``prefetch`` on DuckDB.
 
     Returns:
         The connection, to allow chaining.
@@ -115,4 +140,9 @@ def register(
     # The connection type is erased by the runtime dispatch; every adapter
     # returns the connection it was given.
     adapter: Any = get_adapter(con)
-    return cast(ConT, adapter.register(con, name, ds, chunks=chunks, **kwargs))
+    return cast(
+        ConT,
+        adapter.register(
+            con, name, ds, chunks=chunks, table_names=table_names, **kwargs
+        ),
+    )
--- a/xarray_sql/backends/datafusion.py
+++ b/xarray_sql/backends/datafusion.py
@@ -13,8 +13,15 @@ from typing import Any, TypeGuard
 
 import xarray as xr
 from datafusion import SessionContext
+from datafusion.catalog import Schema
 
-from ..df import Chunks
+from ..df import (
+    Chunks,
+    TableNames,
+    group_vars_by_dims,
+    resolve_table_names,
+    shared_coord_arrays,
+)
 from ..reader import read_xarray_table
 from ..sql import XarrayContext
 from .base import register_adapter
@@ -35,12 +42,37 @@ class DataFusionAdapter:
         ds: xr.Dataset,
         *,
         chunks: Chunks = None,
+        table_names: TableNames = None,
         **kwargs: Any,
     ) -> SessionContext:
-        # XarrayContext.from_dataset adds dim-group splitting, cftime UDF
-        # registration, and round-trip metadata tracking on top of the
-        # plain table registration; use it when available.
+        # XarrayContext.from_dataset adds cftime UDF registration and
+        # round-trip metadata tracking on top of the split below; use it
+        # when available.
         if isinstance(con, XarrayContext):
-            return con.from_dataset(name, ds, chunks=chunks, **kwargs)
-        con.register_table(name, read_xarray_table(ds, chunks, **kwargs))
+            return con.from_dataset(
+                name, ds, chunks=chunks, table_names=table_names, **kwargs
+            )
+
+        groups = group_vars_by_dims(ds)
+        names = resolve_table_names(ds, table_names)
+        if len(groups) <= 1:
+            con.register_table(name, read_xarray_table(ds, chunks, **kwargs))
+            return con
+
+        # A mixed-dimension Dataset becomes one table per dimension group
+        # in a schema named after the Dataset, exactly as XarrayContext
+        # registers it — so ``name.group`` is the same SQL either way.
+        coord_arrays = shared_coord_arrays(ds)
+        schema = Schema.memory_schema(con)
+        con.catalog().register_schema(name, schema)
+        for dims, var_names in groups.items():
+            schema.register_table(
+                names[dims],
+                read_xarray_table(
+                    ds[var_names],
+                    chunks,
+                    coord_arrays=coord_arrays,
+                    **kwargs,
+                ),
+            )
         return con
--- a/xarray_sql/backends/duckdb.py
+++ b/xarray_sql/backends/duckdb.py
@@ -53,12 +53,8 @@ def _mirror_as_schema(
 ) -> None:
     """Expose flat tables as views in a DuckDB schema named *name*.
 
-    ``con.register`` can only place an object in DuckDB's temporary
-    namespace, so a dim-group split lands as flat ``name_group`` tables.
-    A schema of views over them makes the dotted ``name.group`` spelling
-    resolve too — the same SQL text that queries the split on
-    DataFusion, which is the point: the table names are the user's, and
-    they should not have to rewrite queries to change engine.
+    This translates a view to `name_group` can be queried as `name.group`
+    in DuckDB.
 
     *tables* maps each group's table name to the flat name it was
     registered under.
--- a/xarray_sql/backends/duckdb.py
+++ b/xarray_sql/backends/duckdb.py
@@ -21,11 +21,18 @@ the labeled round-trip.
 
 from __future__ import annotations
 
+import warnings
 from typing import TYPE_CHECKING, Any, TypeGuard
 
 import xarray as xr
 
-from ..df import Chunks, group_vars_by_dims
+from ..df import (
+    Chunks,
+    TableNames,
+    group_vars_by_dims,
+    resolve_table_names,
+    shared_coord_arrays,
+)
 from .base import register_adapter
 from .pyarrow import XarrayArrowStream, XarrayPushdownDataset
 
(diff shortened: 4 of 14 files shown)


xarray_sql/backends/duckdb.py
```diff
@@ -35,6 +42,44 @@
...
+    """Expose flat tables as views in a DuckDB schema named *name*.
+
+    This translates a view to `name_group` can be queried as `name.group`
+    in DuckDB.
+
+    *tables* maps each group's table name to the flat name it was
+    registered under.
+    """
+    try:
+        con.execute(f"CREATE SCHEMA IF NOT EXISTS {_quote(name)}")
```
ghostiee-11, line 95: con.register only puts things in the temp namespace but this schema is durable. so on a file backed db the views stick around after the session and point at nothing. reopened one and got `Table with name era5_surface does not exist!`

tried `CREATE SCHEMA temp.era5` and duckdb won't allow it, so maybe only mirror on in-memory connections? or at least mention it in limitations.md

xarray_sql/backends/datafusion.py
```diff
@@ -35,12 +42,37 @@ def register(
...
+                name, ds, chunks=chunks, table_names=table_names, **kwargs
+            )
+
+        groups = group_vars_by_dims(ds)
+        names = resolve_table_names(ds, table_names)
+        if len(groups) <= 1:
+            con.register_table(name, read_xarray_table(ds, chunks, **kwargs))
+            return con
+
+        # A mixed-dimension Dataset becomes one table per dimension group
```
ghostiee-11, line 62: the old path did register a mixed dim dataset as one table. on a 4x3 / 4x2x3 fixture i get `COUNT(*)` = 24 with t2m broadcast across level, so it worked, just weirdly.

no objection to splitting, but it's a break and nothing records it. no test on the old shape and changelog is untouched

xarray_sql/backends/pyarrow.py
```diff
@@ -1141,3 +1145,82 @@ def arrow_dataset(
...
+
+    def table_name(dims: tuple[str, ...]) -> str:
+        group = names[dims]
+        return f"{name}_{group}" if name else group
+
+    if len(groups) <= 1:
+        # One group is one table, named `name` — there is no group to
+        # tell apart. Without a `name`, it takes the group's own.
+        only = name if name else next(iter(names.values()), "scalar")
+        return {only: arrow_dataset(ds, chunks, **kwargs)}
```
ghostiee-11: this branch calls arrow_dataset and the other builds XarrayPushdownDataset, so which kwargs work depends on the data. `arrow_datasets(uniform_ds, "x", _iteration_callback=f)` is a TypeError, the mixed one is fine.

using XarrayPushdownDataset here too would line them up

xarray_sql/backends/pyarrow.py
```diff
@@ -1141,3 +1145,82 @@ def arrow_dataset(
...
+            chunks.
+        table_names: Maps a dimension group's exact dim tuple to the name
+            its table takes, e.g.
+            ``{('time', 'latitude', 'longitude'): 'surface'}``. Groups
+            left unnamed take their dimensions joined by underscores
+            (``time_latitude_longitude``); the group holding scalar
+            variables, if any, takes ``scalar``.
+        **kwargs: Forwarded to
+            [arrow_dataset][xarray_sql.backends.pyarrow.arrow_dataset]
+            (``batch_size``, ``prefetch``, ``geometry``, ...), applied to
```
ghostiee-11: since geometry goes to every group it can never work on a mixed dim dataset. `geometry dims ['lat', 'lon'] are not columns of the table; available: ['proj']`

skip the groups that don't have those dims, or say here that geometry needs a uniform dataset

xarray_sql/backends/duckdb.py
```diff
@@ -35,6 +42,44 @@
...
+        for group, flat in tables.items():
+            con.execute(
+                f"CREATE OR REPLACE VIEW {_quote(name)}.{_quote(group)} "
```
Mmoncadaisla, line 113: Mmm I'm worried that this could silently replace an existing view in the duckdb database, including one that wasn't created by `xarray-sql`. Should replacing an existing view require explicit user opt-in?

Perhaps we could raise an error by default when the target view already exists, and support an option such as `overwrite=True`? When enabled, we could also warn that an existing view was replaced.

Something along these lines:

```python
def mirror_as_schema(con, schema, tables, *, overwrite=False):
    existing = find_existing_views(con, schema)
    conflicts = set(tables) & existing

    if conflicts and not overwrite:
        raise ValueError(
            f"Views already exist: {sorted(conflicts)}. "
            "Pass overwrite=True to replace them."
        )

    with transaction(con):
        ensure_schema(con, schema)
        for group, source in tables.items():
            create_view(
                con, schema, group, source,
                replace=overwrite,
            )

    if conflicts:
        warnings.warn(
            f"Replaced existing views: {sorted(conflicts)}"
        )
```

WDYT?

xarray_sql/df.py
```diff
@@ -149,6 +151,59 @@ def group_vars_by_dims(ds: xr.Dataset) -> dict[tuple[str, ...], list[str]]:
...
+            would silently register one table over the other.
+    """
+    overrides = table_names or {}
+    names = {
+        dims: overrides.get(dims) or default_table_name(dims)
+        for dims in group_vars_by_dims(ds)
+    }
+    taken: dict[str, tuple[str, ...]] = {}
+    for dims, name in names.items():
+        if name in taken:
```
Mmoncadaisla: This check is case sensitive, but (at least) duckdb treats identifiers as case insensitive, even when double-quoted, as opposed to postgres for instance, see https://duckdb.org/docs/current/sql/dialect/keywords_and_identifiers#case-sensitivity-of-identifiers. 

>Identifiers in DuckDB are always case-insensitive, similarly to PostgreSQL. However, unlike PostgreSQL (and some other major SQL implementations), DuckDB also treats quoted identifiers as case-insensitive.

therefore, a mapping that names two groups surface and SURFACE passes this validation, but the second registration silently replaces the first.

Could we add a (duckdb-specific?) collision check before any `con.register()` calls, using its ASCII case insensitive identifier rules? We could preserve the original spelling for registration, but raise a clear error asking the user to choose distinct names when a collision is detected. 

I'd also think it's worth adding a test to ensure no regressions on this. Also, the overwrite option suggested above should not bypass this

xarray_sql/backends/base.py
```diff
@@ -18,7 +18,7 @@
```
Mmoncadaisla, line 18: Would it make sense to log a summary of the tables exposed after registration? I'm thinking that since one dataset can now produce several tables, something like Registered dataset
'era5': era5.surface, era5.atmosphere would help users discover what they can query.

Mmoncadaisla (changes requested):
thanks a lot for this! we're closer to the multi-engine release 🚀 I just added a few comments, nothing too critical though

xarray_sql/backends/duckdb.py
alxmrs: 🤖 Confirmed — reproduced this exactly (create schema+view on a writable file-backed DB, close, reopen: `Table with name era5_surface does not exist!`). Fixed in 4b2f017: `_mirror_as_schema` now checks whether the connection is memory-backed (`PRAGMA database_list`) and skips creating the view on *any* file-backed connection, read-only or not, warning and leaving the flat tables instead. Generalized the `limitations.md`/`engines.md` notes to describe the in-memory-only rule. Added a regression test (`test_file_backed_duckdb_warns_and_skips_the_dangling_mirror`).

xarray_sql/backends/datafusion.py
alxmrs: 🤖 Agreed this is a behavior change, but I don't think it's one worth preserving: the old single-table registration was silently broadcast-joining `t2m` against `temperature`, producing a plausible-looking row count that was actually wrong — a correctness bug, not a documented feature. Added `test_plain_session_context_no_longer_flattens_mixed_dims` to pin that querying the old flattened shape now fails loudly instead of returning a wrong number.

Separately: the PR description says `CHANGELOG.md` is untouched, but there's no `CHANGELOG.md` anywhere in this repo's git history — that line looks like it was hallucinated when the PR body was drafted. Flagging for @alxmrs to decide whether to start one, since I didn't want to invent that convention unilaterally.

xarray_sql/backends/pyarrow.py
alxmrs: 🤖 Confirmed and fixed in 4b2f017 — the single-group branch now builds `XarrayPushdownDataset` directly instead of going through `arrow_dataset()`'s narrower fixed signature, so `_iteration_callback` (and anything else the class accepts) works the same regardless of whether the Dataset happens to be uniform. Added `test_arrow_datasets_accepts_the_same_kwargs_either_shape`.

xarray_sql/backends/pyarrow.py
alxmrs: 🤖 Fixed as you suggested — `geometry=` now only reaches the dimension groups whose dims actually include both, other groups come back without a geometry column instead of raising. Added `test_arrow_datasets_geometry_skips_groups_without_the_dims` and a `limitations.md` note.

xarray_sql/backends/duckdb.py
alxmrs: 🤖 Implemented a narrower version of this rather than the full `overwrite=` API you sketched: `_mirror_as_schema` now checks `duckdb_tables()` for the target schema and refuses (warns, doesn't crash — flat tables still register) only when a **real table** already occupies that name. Re-registering over xarray-sql's own views — the common "re-run this cell" case — still just works without needing a flag on every call.

Reasoning: the actual data-loss risk is specifically "a view silently replaces a table you made yourself," and that's now hard-blocked. A view replacing an *earlier view* from a prior registration seemed fine to keep silent, since that's what re-registration is supposed to do, and requiring `overwrite=True` on every call for that common case felt like the wrong ergonomic tradeoff for the rarer risk. Happy to add the stricter opt-in if you feel the view-over-view case should also require an explicit flag. Added `test_duckdb_refuses_to_mirror_over_a_real_table`.

xarray_sql/df.py
alxmrs: 🤖 Good catch, confirmed and fixed. Added a `case_insensitive` option to `resolve_table_names`, used only by the DuckDB adapter — kept case-sensitive for DataFusion/pyarrow since they don't share this constraint and I didn't want to impose DuckDB's identifier rules on engines that don't need them. Preserves the original spelling for registration and raises before any `con.register()` call happens. Added `test_duckdb_rejects_case_insensitive_name_collision`. The real-table overwrite check from the comment above runs independently and isn't bypassed by this.

xarray_sql/backends/base.py
alxmrs: 🤖 Pushing back on this one rather than implementing it: there's no `logging` usage anywhere in this codebase today — every user-facing runtime notice goes through `warnings.warn` (see the file-backed-connection case above), so adding default logging here would be a new, inconsistent pattern, and would be noisy for callers registering many datasets in a loop. Discoverability is already available natively per engine without duplicating it — `duckdb_views()`/`duckdb_tables()` on DuckDB, the catalog on DataFusion — which matches the library's own stated philosophy of translating data rather than re-implementing engine features. Open to revisiting if you feel strongly, but wanted to explain the reasoning rather than silently skip it.

docs/limitations.md
```diff
@@ -144,12 +144,28 @@ A materialized table or bare C-stream has no query behind it, so the
 re-execution form of `chunks=` cannot serve it — `spill=True` (one
 pass to a temporary Parquet file) is the chunked path for these.
 
-### Mixed-dimension datasets split into one DuckDB table per dim group
-
-DuckDB registration has no schema namespace, so variables with
-different dims land in suffixed tables (`<name>_<dims>`), sharing one
-set of coordinate reads. DataFusion registers the same layout as
-`name.group` tables inside one schema.
+### A file-backed DuckDB connection loses the dotted table names
```
alxmrs, line 147: I'm curious about how this limitation works in practice. Xarray may make it appear to duckdb that the data is only in memory even if it is actually backed by files locally. Xarray would also handle the chunking and streaming similarly to how duckdb would. 

In a future PR, we should empirically test this limitation.

Edit: disclaimer: I don't know too much about duckdb.

xarray_sql/backends/base.py
alxmrs: Yeah `show tables` is preferred for this- users should find this out via SQL.

xarray_sql/backends/datafusion.py
alxmrs: I agree we should pursue correctness, not politeness. No need to start a changelog file, people can review the gh release notes. 

Thansk Aman for catching this issue, it should be addressed.

xarray_sql/backends/duckdb.py
```diff
@@ -35,6 +42,91 @@
 __all__ = ["DuckDBAdapter", "XarrayArrowStream", "XarrayPushdownDataset"]
 
 
+def _quote(identifier: str) -> str:
+    """Render *identifier* as a quoted SQL identifier."""
+    escaped = identifier.replace('"', '""')
+    return f'"{escaped}"'
+
+
+def _is_in_memory(con: duckdb.DuckDBPyConnection) -> bool:
```
alxmrs, line 51: Idea: Should we use a contextmanager to manage this read only table lifetime?

xarray_sql/df.py
```diff
@@ -149,6 +151,80 @@ def group_vars_by_dims(ds: xr.Dataset) -> dict[tuple[str, ...], list[str]]:
...
+
+    Raises:
+        ValueError: if two groups would end up with the same name (or,
+            with ``case_insensitive=True``, names differing only in
+            case), which would silently register one table over the
+            other.
+    """
+    overrides = table_names or {}
+    names = {
+        dims: overrides.get(dims) or default_table_name(dims)
```
alxmrs: Optional Nit: get has a second argument, which is the default value if the value at key dims is not present.

alxmrs (commented):
A few more thoughts from me.

docs/limitations.md
alxmrs: 🤖 Good thing to nail down — but I don't think there's ambiguity here, since these are two independent "in memory" questions: whether *xarray's data* is in memory, and whether *DuckDB's own database* is in memory.

The check (`PRAGMA database_list`) only asks the second question. `con.register()` always creates the same kind of binding — a client-side Python object reference held by the `DuckDBPyConnection` — regardless of whether the xarray Dataset behind it is backed by local NetCDF, remote Zarr, or plain numpy arrays. That binding's lifetime is tied to the *connection object*, never to the data source. What decides whether the mirrored view dangles is only whether the *connection* has an on-disk catalog for that view to persist into (`duckdb.connect("x.db")`) or not (`duckdb.connect()`) — nothing about xarray's chunking or streaming enters into it.

I actually did empirically verify this before writing the fix (reproduced the exact dangling-view error with a plain pyarrow table, no xarray involved: register on a file-backed connection, create the view, close, reopen → `Table with name ... does not exist!`), and `test_file_backed_duckdb_warns_and_skips_the_dangling_mirror` pins that same scenario end-to-end with a real Dataset, so I'd consider the "empirically test this" ask already covered in this PR rather than deferred. Added a clarifying paragraph to `limitations.md` in 6cbb5a9 to make this distinction explicit for future readers, since it's an easy thing to conflate.

xarray_sql/backends/base.py
alxmrs: 🤖 👍 leaving it as-is then, no code change needed.

xarray_sql/backends/datafusion.py
alxmrs: 🤖 Sounds good, no changelog file added. Regression test (`test_plain_session_context_no_longer_flattens_mixed_dims`) is already in from the previous round, pinning that the old broadcast is gone rather than just documented.

xarray_sql/df.py
alxmrs: 🤖 Applied in 6cbb5a9 — `overrides.get(dims, default_table_name(dims))`, dropping the `or`. One behavior note for the record: this makes an explicit empty-string override (`table_names={dims: ''}`) get used literally instead of silently falling back to the default name — but that was never a meaningful case (an empty table name isn't valid SQL anyway), so I don't think it's worth guarding against.

xarray_sql/backends/duckdb.py
alxmrs: 🤖 I want to make sure I build the right thing here rather than guess — a couple of readings of "contextmanager for the read-only table lifetime" seem plausible to me, and they lead to pretty different implementations:

(a) Tie the *mirrored view's* lifetime to a `with` block, so it gets explicitly dropped (`DROP VIEW ...`) on exit instead of relying on detection to avoid creating it at all — e.g. `with xql.register(con, "era5", ds, table_names=...) as con:` drops the views when the block exits.

(b) Something narrower about the read-only-connection path specifically (this function is currently used for both the read-only and the writable-but-persistent case) — e.g. wrap the schema-creation attempt itself in a context manager for cleanup/rollback on failure, rather than about outliving the Python session.

If it's (a): I'd lean against it as the *default* behavior, for a reason beyond just "it's more code" — it only protects the happy path. A crash, `kill -9`, or a notebook cell that errors before `__exit__` runs all still leave the dangling view, and DuckDB's catalog is durable specifically so it survives exactly those cases. Detecting file-backed connections up front and never creating the artifact is the version that's correct regardless of how the process ends. That said, it could be a reasonable **opt-in** for someone who wants the dotted spelling on a persistent DB for the duration of one script and is fine with best-effort cleanup — happy to add it as an explicit `with xql.mirrored(con, ...):`-style helper if that's useful, just didn't want to default to something that only looks safe.

If it's (b), or something else — can you say more about what you had in mind?

alxmrs: Discussed in the chat context. Not relevant as originally thought.

Mmoncadaisla (approved):
LGTM, thank you, Al!! 🙌🏼

docs/limitations.md
Mmoncadaisla: I wouldn't slow this PR down because of this, we can create an issue for further exploration of this behaviour and tackle it in the future if needed. I look forward to seeing duckdb users first and prioritise future developments based on their feedback

